PR #200, opened 2016-07-12
ig-sw

apurvaraman:
C++ frontend for ighutil.


matsen:
@apurvaraman and Anna: I think that by removing ighutil we eliminate the cython dependency (see https://github.com/psathyrella/partis/blob/master/Dockerfile#L14). Can you remove it and verify that this works? If so, commit that removal from this PR.

Thanks!

psathyrella:
dear (near!) future self: also remove ighutil from deps and manual and whatnot.

PR #216, opened 2016-10-03
Clarifying the annotations from partitioning

krdav:


psathyrella:
Awesome! looks good, thanks.

PR #295, opened 2019-08-29
process linearham annotations

eharkins:


python/utils.py
```diff
@@ -784,22 +787,23 @@ def write_airr_output(outfname, annotation_list, cpath, failed_queries, debug=Fa
                 writer.writerow({'sequence_id' : failfo['unique_ids'][0], 'sequence' : failfo['input_seqs'][0]})
 
 # ----------------------------------------------------------------------------------------
-def get_linearham_line(lh_line, seqfos, glfo, debug=False):
-    line = {}
-    for lhk in lh_line:  # [k for k in lh_line if k in linearham_headers and linearham_headers[k] is not None]:  # limit to the ones with a direct partis correspondence
-        if lhk not in linearham_headers or linearham_headers[lhk] is None:
+def process_input_linearham_line(lh_line):
+    """ convert <lh_line> (see linearham_headers). Modifies line. """
```
psathyrella: suggest "Modifies <lh_line."

Also, if it's modifying it, it might be clearer to not also return it?

python/utils.py
```diff
@@ -784,22 +787,23 @@ def write_airr_output(outfname, annotation_list, cpath, failed_queries, debug=Fa
...
+    for lhk in set(lh_line):
```
psathyrella: I don't think I understand what the set() is for -- dict keys already have to be unique, right?

python/utils.py
```diff
@@ -784,22 +787,23 @@ def write_airr_output(outfname, annotation_list, cpath, failed_queries, debug=Fa
...
-    line['indelfos'] = [indelutils.get_empty_indel() for _ in seqfos]
-    # line['dj_insertion'] = 'GGGGGGGGGGGGGGGGGGGGG'  # TODO
-    add_implicit_info(glfo, line, check_line_keys=True)  # TODO remove check_line_keys
+        lh_line[linearham_headers[lhk]] = lh_line[lhk]
+        del lh_line[lhk] #remove lh_line keys once corresponding linearham_headers key added
+    process_input_line(lh_line)
+    return lh_line
 
+# ----------------------------------------------------------------------------------------
+def update_line(line, updates, glfo, debug=False):
```
psathyrella: Can we call this something more evocative that "update"? In the context of the dict function .update() it's clear, but anyone just looking through fcn signatures I think won't have a clue what it does. Hm, actually, the more I think about it, I think it's hard to come up with a name because what it's doing is a super special case to linearham. I.e. overwrite-keys-in-line-with-those-from-update-if-they're-there-otherwise-add-them is something that I'm pretty confident we'd never do in another context. Can this maybe not be a fcn in utils and instead go in the calling code?

python/utils.py
eharkins: @psathyrella since we modify the dict keys while iterating. Is this bad style / bad hack?

psathyrella: oh, duh, right, no i just didn't notice. Good to add the comment though.

PR #296, opened 2019-09-17
utils.get_linearham_bounds comments, debug statements, small changes

eharkins:
Comments, changes, etc added while testing sw-overlap stuff including [ig-sw bug fix](https://github.com/psathyrella/partis/commit/471e5eac6d2b0fbdbb2b6024c81af14cdc3d9399) and [overlap resolution](https://github.com/psathyrella/partis/commit/bf4452b6b62fdbde5130ebcaab50d95c15994364).

### DIFF
--- a/python/utils.py
+++ b/python/utils.py
@@ -331,8 +331,6 @@ extra_annotation_headers = [  # you can specify additional columns (that you wan
     'linearham-info',
 ] + list(implicit_linekeys)  # NOTE some of the ones in <implicit_linekeys> are already in <annotation_headers>
 
-null_linearham_info = {'flexbounds' : None, 'relpos' : None}
-
 linekeys['extra'] = extra_annotation_headers
 all_linekeys = set([k for cols in linekeys.values() for k in cols])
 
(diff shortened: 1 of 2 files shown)


python/utils.py
```diff
@@ -1429,12 +1431,16 @@ def re_sort_per_gene_support(line):
             line[region + '_per_gene_support'] = collections.OrderedDict(sorted(line[region + '_per_gene_support'].items(), key=operator.itemgetter(1), reverse=True))
 
 # ----------------------------------------------------------------------------------------
-def add_linearham_info(sw_info, annotation_list, debug=False):
+def add_linearham_info(sw_info, annotation_list, min_cluster_size=None, debug=False):
     n_already_there = 0
     for line in annotation_list:
+        if min_cluster_size is not None and len(line['unique_ids']) < min_cluster_size:
+            print '       %s adding null linearham info to line: %s, because cluster is less than the passed <min_cluster_size> value of %d' % (color('yellow', 'warning'), ':'.join(line['unique_ids']), min_cluster_size)
+            line['linearham-info'] = null_linearham_info
```
psathyrella: I think it would be safer to either use a copy.deepcopy() here, or have a function return it (a la [this](https://github.com/psathyrella/partis/blob/dev/python/indelutils.py#L11)). The way it is now, if this line's linearham info gets subsequently modified, i think it will also modify the linearham info of all other lines that have null info.

eharkins: Ah, you're right. Good catch, thanks!

Issue #52, opened 2015-01-11
Harlan says he has a way of doing high-throughput antigen identification


psathyrella:
chat about what that means

PR #341, opened 2026-03-11
Phase 1: Disjoint grouping core implementation

s-canchi:
## Summary

- Three independently callable subcommands: `disjoint-group`, per-group `partition`, `assemble-groups`
- YAML manifest as handoff between steps
- Works for paired (multi-locus) and unpaired (single-locus) data
- Works for simulated and real data
- Disjoint grouping does not degrade clustering accuracy (validated HA and vsearch against standard partition)
- CDR3 length disjointness validated on simulated and real data
- Test framework integrated into `test/test.py` with reference results for all four configs
- No regressions (`partis-test.py --quick` and `--paired --no-simu` pass)

## Test plan

- [x] `partis-test.py --quick` passes
- [x] `partis-test.py --paired --no-simu` passes
- [x] Disjoint grouping tests pass for paired simu, paired data, unpaired simu, unpaired data
- [x] Clustering accuracy comparison shows no degradation

Related: #337


s-canchi:
#### Design choices in Phase 1

1. Independent subcommands as internal machinery: three separate steps (`disjoint-group`, `partition`, `assemble-groups`) rather than a single command. A `disjoint-partition` wrapper for standard partis single-command UX is planned for the next phase.

2. YAML manifest as handoff contract: new artifact type for partis, describing groups, file paths, and metadata between steps.

3. Per-group FASTA files as intermediate artifacts: writes to disk between steps rather than keeping data in memory, enabling independent parallel execution. I/O metrics to be captured during scale validation to quantify the tradeoff.

4. Custom merge logic instead of `merge_paired_yamls()`: `merge_yamls()` asserts matching partition step counts, which independent runs do not produce. Custom logic extracts only the best partition from each group and reconciles germlines.

psathyrella:
### testing
  - Should run on --slow tests, as well as paired paper validation data, to check performance
  - disjoint tests seem not to work with --dry? I use --dry quite a bit to get the commands that would run to help with testing other things
  - it appears that the disjoint-group results from testing aren't actually compared? I don't see them in the comparison output
  - disjoint group tests should be in same order as other tests, as a variation of partition (probably after subset-partition)
  - what's the purpose of having --disjoint-dir separate from --workdir?
  - it might be better to test disjoint-group without --parameter-dir since in practice I think it would always infer its own parameters
  - it seems to rerun group creation even if the groups are already there, which seems worth avoiding. 
  - it prints step 1: and step 2: but not 3: (just the component partition commands)
  - I don't think I understand why we need the new function  run_disjoint_group_tests() -- it doesn't fit with the existing tests, which all consist of single commands that run within a common framework. It also seems like most of the code in this new function would need to be duplicated elsewhere if you were running disjoint partitioning on non-test data.
  - i don't think we want a check for families spread across disjoint groups -- this will just fail if the shm indels fall within cdr3, it's just an assumption. We want to run the normal testing performance comparison on results  

### disjointgrouper.py
  - do we really need a new function to group seqs by cdr3? I think this is done in several other places already (e.g. paircluster.py 1766; see also utils.collapse_naive_seqs_with_hashes and utils.split_clusters_by_cdr3). Even if we have to modify an existing function, it would be nice to not start from scratch
  - it would be nice to use utils.write_fasta e.g. in write_group_fastas
  - why is there a hardcoded yaml version in write_manifest? 
  - get_sw_cache_path: I would really prefer if we could sort out which location we expect, rather than just looking in both, since (for instance) the latter is prone to using the wrong one if they both exist. It would probably also be simpler to follow subset-partition in handling everything as paired samples since it's so trivial to rearrange unpaired input to behave as paired input. 
  - get_loci: if we really want this function it should probably be in utils.py since that's where the functions with similar names live. But I think really it would be better to not handle paired and unpaired as separate logic paths
  - find_partis_cmd: this also seems like, if we really do need it, it should be a general thing that lots of commands need, so in utils.py. But the docs just say that partis should now be in the path, so is there a reason we can't just continue that convention?
  - I feel the same way about get_infname_for_locus: it even has an exception that it isn't sure what to use. I'm quite skeptical that we need any of this guessing infrastructure
  - get_parameter_dir_for_locus: samesies
  - auto_cache_parameters: it seems like this passes n-procs and is-simu, but what about all the other args that could be set that need to be passed to cache-parameters? See partis.run_step, which is doing the same thing. Hopefully we can use that function rather than starting a new one?
  - it seems that it's set up to handle all disjoint groups from all loci together, but I'm not sure this is best. It seems it might be a lot simpler to handle one locus at a time without needing to know anything about loci. 
  - it needs to run locus splitting/pair info extraction so we don't lose that functionality that exists elsewhere. 
  - assemble_merged_output: it seems an awful lot simpler to me to add an option to merge_yamls rather than duplicate the entire function to avoid an assertion.   
  - this doesn't seem to do anything to handle pairing info, e.g. merging paired partition, pair info cleaning. So unless I'm misunderstanding (I could be!) this is essentially only handling the single chain partition step of paired clustering -- is this intentional, we'll just add it later?
 
Big picture, sorry in advance if this is supposed to be just a stepping stone, but I think disjoint grouping just needs to be run as an option, eventually default, for partitioning -- we can't require the user to run multiple commands with complicated path dependencies (again, sorry if I'm misunderstanding). Like for instance

```
partis subset-partition --infname xxx.fa --paired-loci --disjoint-group
```
from the user's perspective should run the same as without disjoint grouping, it just runs a more complicated apportionment algorithm with extra parameter caching step. Now, exactly how much of the existing subset partition function should it use? That's for sure still to be decided, they are doing somewhat different, we of cousre don't have to reuse everything. But at minimum, conceptually, this should be behaving similarly.


Again, thanks so much for pushing on this!

s-canchi:
Thank you for the detailed review. I agree with the code-level feedback across the board (reuse existing CDR3 grouping functions, use utils.write_fasta, integrate via run_step(), modify merge_yamls() instead of duplicating, restructure tests into the standard framework, etc). I will follow up with implementation.

On the bigger architectural question, I agree that `subset-partition --disjoint-group` is the right interface rather than separate subcommands.

**What I plan to change in subset-partition**

I will modify `subset_partition()` in `bin/partis` to support a `--disjoint-group` flag. When set, three things change relative to the current random-split behavior:

1. **Apportionment**: instead of loading all sequences into memory and splitting randomly via `subset_paired_queries()`, the disjoint path runs `cache-parameters --only-smith-waterman` as a subprocess (which exits and releases memory), then reads the SW annotations and groups UIDs by cdr3_length. `--parameter-dir` is the escape hatch for cases where SW annotation was already done externally. For the grouping itself, I will reuse existing CDR3 grouping functions. Per-group FASTAs and a manifest are written to disk.

2. **Per-group partition dispatch**: each CDR3 group runs standard partition as a subprocess, same as current subset runs. The manifest enables resumability (skip groups whose output already exists).

3. **Merge step**: since CDR3 groups are biologically disjoint (no clonal family spans two CDR3 lengths), the merge is concatenation + germline reconciliation via `merge_yamls()` with the partition count assertion relaxed. The re-partition step that random-split subset-partition requires (because random subsets can split families) is not needed.

Without the `--disjoint-group` flag, existing subset-partition behavior is completely untouched.

**Paired path**: everything routes through the existing paired path (unpaired as degenerate single locus), using `run_all_loci()` / `run_step()` for parameter caching and locus handling.

**HPC / resource optimization**

One motivation for the original separate-command design was that each step has different resource profiles. This applies to SW annotation too, since for multi-million sequence datasets chunking SW annotation across nodes is much faster than running it as a single serial process. Would you be open to a mode where the disjoint path writes the manifest/groups and exits, so an external orchestrator can submit per-group jobs with right-sized resources? This could be a flag like `--manifest-only`, or I could keep the individual subcommands available as building blocks. The single-command path would remain the default.

**Pairing info**

Phase 1 scope was single-chain partition correctness. The paired pairing/cleaning steps should work unchanged on the merged single-chain outputs from disjoint grouping, but I have not confirmed that yet. I will verify as part of the integration.

psathyrella:
This sounds great, thanks!


> **What I plan to change in subset-partition**
> 
> I will modify `subset_partition()` in `bin/partis` to support a `--disjoint-group` flag. When set, three things change relative to the current random-split behavior:
> 
>     1. **Apportionment**: instead of loading all sequences into memory and splitting randomly via `subset_paired_queries()`, the disjoint path runs `cache-parameters --only-smith-waterman` as a subprocess (which exits and releases memory), then reads the SW annotations and groups UIDs by cdr3_length. `--parameter-dir` is the escape hatch for cases where SW annotation was already done externally. For the grouping itself, I will reuse existing CDR3 grouping functions. Per-group FASTAs and a manifest are written to disk.

great. 

>     2. **Per-group partition dispatch**: each CDR3 group runs standard partition as a subprocess, same as current subset runs. The manifest enables resumability (skip groups whose output already exists).

This sounds good, although the current subset partition already skips individual subset partition runs if output exists, although using the manifest does sound like a nice addition. 
 
>     3. **Merge step**: since CDR3 groups are biologically disjoint (no clonal family spans two CDR3 lengths), the merge is concatenation + germline reconciliation via `merge_yamls()` with the partition count assertion relaxed. The re-partition step that random-split subset-partition requires (because random subsets can split families) is not needed.

yes, exactly, just concatenation without looking for new merges.

> Without the `--disjoint-group` flag, existing subset-partition behavior is completely untouched.
> 
> **Paired path**: everything routes through the existing paired path (unpaired as degenerate single locus), using `run_all_loci()` / `run_step()` for parameter caching and locus handling.
> 
> **HPC / resource optimization**
> 
> One motivation for the original separate-command design was that each step has different resource profiles. This applies to SW annotation too, since for multi-million sequence datasets chunking SW annotation across nodes is much faster than running it as a single serial process. Would you be open to a mode where the disjoint path writes the manifest/groups and exits, so an external orchestrator can submit per-group jobs with right-sized resources? This could be a flag like `--manifest-only`, or I could keep the individual subcommands available as building blocks. The single-command path would remain the default.

Yes, that sounds great. I think we want to enable the case of super large samples that require finagling different resource allocations at different steps, but the most common use case is still running on one machine with one command, so that should stay the default.

s-canchi:
#### Post-review update: all items addressed + validation results

All review feedback has been addressed and validation is complete across simulated tests and the paired paper datasets.

Architecture (big picture)
- Integrated disjoint grouping as `subset-partition --disjoint-group` instead of separate subcommands. From the user perspective it runs the same as standard subset-partition, just with a different apportionment algorithm.
- Everything runs through `subset_partition()` in `bin/partis`. The disjoint path replaces the random-split step with: SW annotation + CDR3 length grouping + per-group partition + assembly merge.
- Locus splitting and pair info extraction handled by existing `run_all_loci()` framework.

disjointgrouper.py cleanup
- Removed 6 duplicate helper functions (~80 lines): `find_partis_cmd()`, `get_sw_cache_path()`, `get_loci()`, `get_infname_for_locus()`, `get_parameter_dir_for_locus()`, `auto_cache_parameters()`.
- Replaced custom CDR3 grouping with existing `utils.group_seqs_by_value()`.
- Using `utils.write_fasta()` instead of manual FASTA writing.
- Removed hardcoded YAML version in `write_manifest()`.
- Modified `merge_yamls()` with `best_partition_only` parameter instead of duplicating merge logic in `assemble_merged_output()`.

Test restructuring
- Removed `run_disjoint_group_tests()`. Disjoint test is now a standard `self.tests` entry processed by `run()`/`compare_stuff()`.
- `--dry` works automatically through the standard framework.
- Results are compared via CCF metrics (purity/completeness) and pair-clean performance, same as all other partition tests.
- Removed the disjointness assertion (CDR3 families not splitting across groups). As noted in review, this is an assumption, not something to assert. Validation is through the standard CCF metrics.
- Test placed after subset-partition in the test ordering.
- `--disjoint-dir` removed; uses standard `--paired-outdir`/`--outdir` paths.
- Test runs without `--parameter-dir` (SW annotation auto-triggers).
- Skip-if-exists logic added for grouping step, per-group partition, and assembly.
- Step 3 label now prints in test output.

Pair cleaning integration
- `--input-metafnames` passed to per-group partition commands so `paired-uids` propagate through to per-group annotations.
- Assembled output linked to `single-chain/partition-{locus}.yaml` where `combine_inf_chains()` expects it.
- `combine_inf_chains()` called after disjoint subset partition returns in `run_all_loci()`.
- Pair-cleaned output and `true-pair-clean-performance.csv` produced identically to standard partition.

#### Validation results

#### Simulated data: normal paired tests (`--paired --no-simu`, ~56 seqs per locus)

| Method | Purity | Completeness | Pair Clean Correct | Unpaired | Mispaired |
|---|---|---|---|---|---|
| partition | 0.980 | 0.588 | 0.726 | 0.133 | 0.142 |
| subset | 0.980 | 0.677 | 0.726 | 0.133 | 0.142 |
| disjoint | 0.980 | 0.568 | 0.726 | 0.133 | 0.142 |

#### Simulated data: slow paired tests (`--slow --paired`, ~1700 seqs)

| Method | Purity | Completeness | Pair Clean Correct | Unpaired | Mispaired |
|---|---|---|---|---|---|
| partition | 0.890 | 0.788 | 0.795 | 0.156 | 0.049 |
| subset | 0.878 | 0.776 | 0.795 | 0.145 | 0.055 |
| disjoint | 0.903 | 0.792 | 0.800 | 0.159 | 0.042 |

Disjoint partition matches or slightly outperforms standard partition on all metrics at both scales.

#### Paired paper validation datasets (10x data, Zenodo 6998443)

Ran `subset-partition --disjoint-group --paired-loci` on all 4 datasets from the paired paper using existing parameters. Compared pair-cleaned output against existing standard partition results.

Single-chain (pre pair-clean) sequence counts are identical between standard and disjoint for all datasets and loci. Zero sequence loss from CDR3 length grouping.

| Dataset | Locus | Std seqs | DJ seqs | Std purity | Std completeness | DJ purity | DJ completeness |
|---|---|---|---|---|---|---|---|
| hs-2-pbmc | igh | 1007 | 909 | 1.0000 | 1.0000 | 1.0000 | 1.0000 |
| hs-2-pbmc | igk | 620 | 605 | 0.9537 | 0.9653 | 0.9653 | 0.9537 |
| hs-2-pbmc | igl | 426 | 417 | 0.9496 | 0.9616 | 0.9616 | 0.9496 |
| mm-balbc | igh | 923 | 849 | 0.9988 | 1.0000 | 1.0000 | 0.9988 |
| mm-balbc | igk | 935 | 900 | 0.9122 | 0.9589 | 0.9589 | 0.9122 |
| mm-balbc | igl | 61 | 59 | 0.9492 | 0.9492 | 0.9492 | 0.9492 |
| hs-1-postvax | igh | 8498 | 7678 | 0.9995 | 0.9991 | 0.9991 | 0.9995 |
| hs-1-postvax | igk | 4800 | 4764 | 0.9163 | 0.9399 | 0.9399 | 0.9163 |
| hs-1-postvax | igl | 4096 | 4056 | 0.9258 | 0.9383 | 0.9383 | 0.9258 |
| hs-1-prevax | igh | 9187 | 8253 | 0.9992 | 0.9978 | 0.9978 | 0.9992 |
| hs-1-prevax | igk | 5329 | 5284 | 0.9093 | 0.9375 | 0.9375 | 0.9093 |
| hs-1-prevax | igl | 4360 | 4308 | 0.9248 | 0.9136 | 0.9248 | 0.9136 |

"Std purity" = purity of standard partition with disjoint as reference. "DJ purity" = purity of disjoint partition with standard as reference. The metrics are symmetric: purity(A vs B) = completeness(B vs A).

Summary:
- IGH (heavy chain): near-perfect agreement across all datasets (purity/completeness > 0.997). CDR3 length grouping has essentially no effect on heavy chain clustering.
- Light chains (IGK/IGL): 91-97% agreement. Differences arise from pair cleaning operating on slightly different cluster boundaries. Neither method is systematically better.
- Sequence count differences in the final pair-cleaned output (~10% fewer IGH in disjoint) are entirely from pair cleaning, not from CDR3 grouping. Single-chain counts are identical.

#### CDR3 length disjointness

Validated across all test scales: 585 simulated families (200 unpaired + 385 paired), 100 real paired sequences, 1,371 real unpaired sequences, and all 4 paper validation datasets (~40K total sequences). Zero families split across CDR3 length groups in all cases.


If these changes look good, the next phase I had planned out was to implement naive hamming fraction sub-grouping within CDR3 length bins which would become necessary if CDR3 length bins are too large at scale. Testing the CDR3 and CDR3 + hfrac grouping on the 4m subsets from previous benchmarking steps and comparing results would let us compare the method and resource usage on real data. One more motivation for testing the hfrac based sub grouping was to check if enabling HA as default for disjoint-grouping would be feasible. What are your thoughts @psathyrella ?

s-canchi:
#### Two open items from the review

**1. `--disjoint-dir` vs `--workdir`**

I kept `--disjoint-dir` separate from `--workdir` because the disjoint path writes its own artifacts (manifest, per-group FASTAs, per-group partition outputs) that do not map to the simulation output structure that `--workdir` controls. In the integrated `subset-partition --disjoint-group` path, the disjoint directory defaults to `{paired-outdir}/disjoint-groups/` and the `--disjoint-dir` arg is only exposed for the standalone `disjoint-group` and `assemble-groups` subcommands. As discussed, these are kept for standalone execution for at-scale runs.

**2. Arg forwarding gap: not using `run_step()`**

I was not able to use `run_step()` for the disjoint path because it is a nested function inside `run_all_loci()` with closure dependencies that make it uncallable from outside. Instead, both `auto_cache_parameters()` (disjointgrouper.py:165) and the per-group partition commands (bin/partis:358) manually build their subprocess commands and only forward a few args (`--n-procs`, `--is-simu`, `--input-metafnames`, `--locus`). This means other user-provided args that affect partition behavior (e.g. `--n-max-queries`, `--random-seed`, `--only-genes`, `--no-indels`) do not propagate to the disjoint path. I left a TODO at disjointgrouper.py:167 noting this. Not propagating full user args is a gap that should be fixed in this PR. Would you prefer I extract `run_step()` into a standalone function, or adopt the same `sys.argv` deep-copy pattern in the disjoint path?

psathyrella:
Looks great, again, thanks!

  - I should have made it clear -- the paired paper samples I was talking about were simulation -- we just don't learn much from data since the samples are so small that partitioning is trivial
  - Yes it sounds great to go ahead with the hamming fraction grouping as long as it seems like it'll help, and doesn't seem like it'll be a big complication
  - Also yes, I think we should work on running HA on larger samples. I think it's really a conversation around adjusting the default thresholds for turning on --fast and similar options. It's possible we should go back to using sw annotations when --fast is on, as well.

1. --disjoint-dir vs --workdir: `because the disjoint path writes its own artifacts (manifest, per-group FASTAs, per-group partition outputs) that do not map to the simulation output structure that --workdir controls.` I don't think I understand this -- --workdir isn't simulation specific, it's where all the non-persistent stuff lives, which is everything if output paths aren't specified. It's nice that --disjoint-dir doesn't bleed through to actions that don't use it, but what I'm worried about is usability not code behavior. It's seems confusing to me to have 90% or whatever of the code base require someone to set --workdir, but use --disjoint-dir for disjoint grouping (although maybe I'm confused about this). I would imagine that the disjoint stuff would go in its own subdir of either the workdir or output dir, depending on what the user has set, in which case it doesn't seem to me like it would interfere with anything
2. Arg forwarding gap: not using run_step(): You'll want to split it out, I don't think there's anyway we could duplicate the functionality correctly even if we wanted to copy however many lines of code.


Haven't had a chance to go through the code yet, sorry, but I'll do it this morning.

psathyrella:
I think we should also remove a lot of files from the reference testing dirs. It may just be the same files I've had in there before,  but with the 30x or whatever from all the cdr3 subsets it's just too much. I'm not sure whether adding to the existing removal infrastructure, or just gitignoring, is better. Probably just the latter

s-canchi:
Thank you for the feedback.

I addressed all three in this PR:
1. Replace `--disjoint-dir` with a subdir of `--workdir`/`--paired-outdir`. Point taken that having a separate arg is confusing when the rest of partis uses `--workdir`.
2. Extract `run_step()` into a standalone function for full arg forwarding to per-group partition commands and `auto_cache_parameters()`.
3. Gitignore the per-group intermediate files (FASTAs, per-group partition YAMLs, logs) in ref results dirs. The test framework only needs the assembled outputs and manifest for comparison.

**Phase 2 sequencing**

I plan to merge Phase 1 (this PR) first, then work on hfrac sub-grouping as Phase 2 on a new branch. Testing CDR3 and CDR3 + hfrac grouping on the 4m subsets from previous benchmarking steps would be Phase 3 for real evaluation. HA tuning is independent of these goals.

**Paired paper datasets**

When you said the paired paper samples should be run on simulation, do you mean I should simulate data using the per-sample parameters (from the zenodo tarball) at a larger scale, then run disjoint-partition on that simulated data and compare against the known ground truth?

psathyrella:
- do we have any way to check that build_single_locus_cmd handles all the args the same as the old function? It would just be quite difficult to write tests to check all combinations of commands on all possible data samples, so we might need this to be a code type check rather than run time. It would just be really bad if, say, an arg got swapped from the removal list and we didn't notice, since it could easily really change behavior. I'm also a bit worried that we've changed the order of doing different things with these args, which I'm pretty sure could change behavior. EDIT oh sorry, I thought this was a shared function. It looks like it just copies run_step? I don't think that's a good idea, I thought we'd discussed extracting a shared function?
  - a previous comment indicated get_sw_cache_path was removed, but I think it's still there?
  - auto_cache_parameters seems to still have the issues mentioned [here](https://github.com/psathyrella/partis/pull/341#issuecomment-4048743853), although now also has a todo indicating it needs merging with the run_step stuff
  - I don't really understand why we set args.disjoint_dir and args.no_merge_output at the top of disjoint_subset_partition, but if there's any way to avoid setting args like this (rather than on command line) that seems better
  - I'm not sure that running partition on each subset sequentially is the way to go if we have a subset for each cdr3 length. Maybe this is fine, but some of the cdr3 subsets are going to be vastly smaller than others, so it might start getting slow. Maybe it'd fine, maybe we should equalize the subset sizes by merging cdr3 groups, maybe we should run the at the same time (not all sequentially).
  - I think I want to update the test refs and a few other code changes that should be independent of you without merging this PR (since I'm also working on a big PR from erick), would it be ok if I merged in a few commits to dev? This would also reduce the size of the test ref stuff you have here, which would make the diff more manageable
  - it seems that disjoint subset partition uses --parameter-dir, whereas non-dijoint does not. My understanding was that we were treating the subsets the same in both cases -- the job for each subset would auto cache its own parameters. If we're not doing that, I think we'll have to think a little harder about how to handle the parameter dir since it's not usually used for paired loci. Don't we want to avoid having to cache parameters on the entire sample (other than the sw cache step to allow subsetting)? One problem with the current setup is if I run it with paired outdir set but no parameter dir, it puts the parameter dir in _output/, which I don't think is what we want 
  - why do we need the new input metafnames handling at the top of subset_partition?
  - this seems like it calls the new disjoint subset partition from inside the old  subset partition function, but doesn't actually use any of the existing code, which is what I was imagining. I could be wrong but it really seems like they're doing very similar things so I thought we were going to just have them more integrated
  - the args.is_simu block that calculates partition performance metrics seems like it duplicates a lot of code that appears elsewhere (it appears that subset_partition does this via combine_inf_chains). I understand that there's a difference in that subset_partition has the final merging step, but this block just has a lot of fiddly stuff that it seems easy to get out of sync

psathyrella:
> **Phase 2 sequencing**
> 
> I plan to merge Phase 1 (this PR) first, then work on hfrac sub-grouping as Phase 2 on a new branch. Testing CDR3 and CDR3 + hfrac grouping on the 4m subsets from previous benchmarking steps would be Phase 3 for real evaluation. HA tuning is independent of these goals.

yes, that sounds good

> **Paired paper datasets**
> 
> When you said the paired paper samples should be run on simulation, do you mean I should simulate data using the per-sample parameters (from the zenodo tarball) at a larger scale, then run disjoint-partition on that simulated data and compare against the known ground truth?

I was meaning to run on the simulation samples from the paired paper, rather than real data samples. I don't think there's any reason to regenerate the samples, this would just involve running a new method (disjoint subset partition) and making the comparison plots with the new line added to e.g. Fig 3 and the associated supplement figures. It may make more sense for me to do this, I'm not sure, but I think we want to focus on getting the overall code architecture how we want it first.

s-canchi:
Thank you for the detailed review. I agree with the architectural direction. The disjoint path should integrate into existing subset-partition machinery rather than running as a parallel implementation.

Here is how I plan to address this:

- Extract `prep_args()` from `run_step()` into a shared standalone function that both `run_step()` and the disjoint path call. `build_single_locus_cmd()` was a copy, not a shared function - that was a mistake.
- Integrate disjoint dispatch into `subset_partition()` reusing the `get_cmd()` pattern and the same dispatch structure as `run_subsets()`, with the merge step using `merge_yamls(best_partition_only=True)` instead of the second-pass merge-partition.
- Align parameter dir handling with subset-partition behavior where per-group commands auto-cache their own parameters rather than requiring a pre-cached `--parameter-dir`.
- Remove the duplicate CCF calculation block and let the standard `combine_inf_chains()` / `run_all_loci()` framework handle it.
- Remove `auto_cache_parameters()` from the integrated path, args injection for `disjoint_dir`/`no_merge_output`, and the custom `input_metafnames` handling.
- For Phase 1, groups will run sequentially with full `--n-procs` each, same as current subset-partition subsets. Concurrent dispatch or group size equalization would be Phase 2 optimization.

On `get_sw_cache_path`: it was changed from raising an exception to returning `None` (for cleaner control flow), but the function itself was not removed since it is still used to locate the SW cache file.

**Dev branch merges**: No conflict from our side. PR #341 targets the `disjoint-grouping` feature branch, not `main` or `dev` directly. The feature branch can merge into `dev` or `main` depending on your preference.

**Paired paper validation**: Understood that this means running disjoint on the existing simulation samples from the paired paper and adding results to Fig 3. If you plan to run this yourself, should that happen after this PR is merged, or as part of the PR review?

psathyrella:
This sounds good, thanks! We're getting closer and closer.

re: sw cache path function, it just really worries me that it seems like it can return different sw cache paths depending what's in the file system. For instance if we rerun with slightly different input file (e.g. removing failed queries as in the gidoni run) the hash in the cache file name will change and it looks like this function will just choose the first one with maybe alphabetical sorting. It also seems dangerous to not enforce whether it looks in the locus subdir or not -- it seems to me that the code should know where to expect the file based on its command line args.  

For now let's plan on me running on the paired samples later. I'll need to rerun some new shared mutation clustering soon as well, anyway.

s-canchi:
#### Architecture refactor: addressing round 2 review feedback

Five commits addressing the architectural concerns from the 2026-03-17 review.

**Shared arg stripping utility (4cb7b343c)**

Extracted `strip_paired_args()` as a shared function that both `build_single_locus_cmd()` and `prep_args()` in `run_step()` call. This is now a single source of truth for which paired/multi-locus args to strip from single-locus subprocess commands. Also fixed `--single-light-locus` being incorrectly stripped as a boolean arg (it takes a value).

**Disjoint flow integrated into subset_partition() (4cb7b343c)**

Removed `disjoint_subset_partition()` as a separate function. The disjoint flow is now inline in `subset_partition()`, with three focused helpers: `run_disjoint_groups()` for per-group dispatch, `calculate_disjoint_ccfs()` for post-merge CCF calculation, and `link_assembled_to_single_chain()` for symlink setup. `args.disjoint_dir` and `args.no_merge_output` injection removed; assembly calls `validate_assembly()` and `assemble_merged_output()` directly.

**input_metafnames scoped to disjoint block (6fbb1acf7)**

Moved the `input_metafnames` default-setting from the top of `subset_partition()` (where it affected both paths) into the `if args.disjoint_group:` block. Non-disjoint subsets do not need it because they spawn full paired subprocesses that discover meta files via `run_all_loci()`/`get_meta_fns()`. Disjoint per-group commands are single-locus and need it explicitly for `paired-uids` to propagate.

**Parameter dir default fixed (8674e7245)**

Fixed `run_disjoint_group()` in disjointgrouper.py to default `--parameter-dir` to `{paired_outdir}/parameters` when `--paired-outdir` is set, matching `getpdir()` behavior in `run_all_loci()`. Previously it fell back to `_output/` regardless of `--paired-outdir`.

**get_sw_cache_path fixed (978307595, 49e14bd71)**

Addressed both concerns: (1) uses `args.paired_loci` to decide whether to look in `{parameter_dir}/{locus}/` or `{parameter_dir}/`, rather than searching both locations; (2) if multiple `sw-cache-{hash}.yaml` files exist (from re-runs with different input, since partis does not clean up old cache files), warns and picks the most recent by mtime. The newest cache file matches the current parameter dir state since SW/HMM parameter dirs are overwritten on re-run.

**Validation**: all tests pass (`--quick`, `--paired --no-simu`). Disjoint partition metrics unchanged.

psathyrella:
> Disjoint per-group commands are single-locus and need it explicitly for paired-uids to propagate.

If disjoint commands are all single locus, how do we get paired partners together in the final step? I guess if we're assembling, they'll get concatenated into the same paired output, but when does paired clustering happen then?  It seems like we'd never be using the pair information. And I was thinking that we wanted to make assembling optional, but disjoint groups don't seem very useful if the user has to go search through all the groups in each locus to find the paired partner for each sequence.

Would it make more sense to make disjoint apportionment operate on paired queries? That would automatically allow much more splitting, since we'd be splitting on two independent cdr3 lengths, which would square the probabilities. It seems like this would also make the subsequent steps easier. I think this also may be why it's proving difficult to integrate disjoint grouping and subset partition (the current approach just runs disjoint grouping from a nested function in the subset partition function, but it's not clear to me why as written it'd be nested inside.) 

For the arg stripping stuff, could we have CC run a test with all combinations of all actions and command line args? Like not actually running the commands, but verifying that the resulting commands are identical. I think just using --dry for paired actions would work. It's just really hard otherwise to make sure it's correct.

strip_paired_args: I think some of these are just args whose effects we don't want on the single locus subprocesses, not necessarily that they don't apply to single locus

input_metafnames scoped to disjoint block: it's hard for me to verify that this isn't needed for plain subset partition without running some tests with a bunch of different arguments, but are we sure get_meta_fns() does exactly the same thing in all cases? Even if it does, I guess it's nice to have a simplification if we can, but if we're not sure it's the same, why would we need to move it to only affect disjoint group?

get_sw_cache_path fixed: it does seem like an improvement, but I still don't understand why we need to guess. What is disjoint grouping doing differently to subset-partition that it can't just know where the sw cache file should be?

s-canchi:
**Paired clustering flow**

My understanding is that pair cleaning happens after assembly, following the same pattern as standard partition: single-locus partition produces per-locus annotations with `paired-uids` (from `--input-metafnames`), assembly merges them per-locus, symlinks go to `single-chain/`, then `combine_inf_chains()` does paired clustering. But I may be missing something about how this interacts with the disjoint grouping step. Please let me know if this does not match your expectation.

On paired apportionment (grouping on both heavy+light CDR3 lengths): I think that is a good idea worth exploring, but it would require reading paired info before grouping, which the current step 1 does not do. I would plan this for Phase 2 alongside hfrac sub-grouping rather than trying to fit it into this PR.

**Arg stripping verification**

Here is the dry-run output showing subprocess commands generated through `strip_paired_args()` + `prep_args()`:

```
$ partis cache-parameters --paired-loci --paired-indir test/paired/ref-results/test/simu \
    --parameter-dir /tmp/test-strip/parameters --is-simu --random-seed 1 --n-procs 4 --dry-run

run bin/partis cache-parameters --locus igh --parameter-dir /tmp/test-strip/parameters/igh \
    --is-simu --random-seed 1 --n-procs 4 --crash-on-duplicate-uids \
    --infname test/paired/ref-results/test/simu/igh.yaml \
    --sw-cachefname /tmp/test-strip/parameters/igh/sw-cache.yaml \
    --input-metafnames test/paired/ref-results/test/simu/meta.yaml
[igk and igl analogous]
```

Per-group partition commands from `build_single_locus_cmd()` (from slow test log):

```
group 0 (cdr3 30, igh): run bin/partis partition \
    --infname .../disjoint-groups/groups/cdr3-30/igh.fa \
    --outfname .../disjoint-groups/groups/cdr3-30/partition-igh.yaml \
    --parameter-dir .../test/parameters/simu/igh \
    --locus igh --dont-write-git-info --n-procs 10
```

I can build a more systematic comparison across all action and arg combinations if that would help. The dry-run approach only covers the paths that actually run, so edge cases with unusual arg combinations would not be caught.

On the naming: you are right that "strip_paired_args" is not fully accurate. Some of the args are not paired-specific but are stripped because their effects are handled at the outer level (like `--dont-calculate-annotations`). I am open to a better name if you have a suggestion.

**input_metafnames**

You were right to question this. I found that moving the default into the disjoint-only block broke the non-disjoint subset-partition path. `run_subsets()` and `get_cmd()` both use `args.input_metafnames` for pair info handling and forwarding. Fixed in `e1fb18759`, the default is back at the top of `subset_partition()` where both paths can use it.

**get_sw_cache_path**

The reason it uses glob instead of constructing the path deterministically is that `run_disjoint_group()` does not read the input file. It reads the SW cache directly to get CDR3 lengths for grouping. The SW cache filename includes a UID hash (`sw-cache-{md5(uids)[:10]}.yaml`) computed from the input UIDs by `PartitionDriver`. To construct that filename, you would need to read the input file to get the UIDs, which defeats the purpose (the whole point is to avoid loading millions of sequences when the cached annotations already have what is needed).

Subset-partition does not have this issue because its subprocesses go through `PartitionDriver`, which reads the input and computes the hash internally. An alternative would be to pass `--sw-cachefname` explicitly from the outer level, similar to how `prep_args()` does it at line 906, but that path computation lives inside `run_all_loci()` closures.

I am not sure if there is a cleaner way to handle this. If you have ideas I am happy to try them.

psathyrella:
> **Paired clustering flow**
> 
> My understanding is that pair cleaning happens after assembly, following the same pattern as standard partition: single-locus partition produces per-locus annotations with `paired-uids` (from `--input-metafnames`), assembly merges them per-locus, symlinks go to `single-chain/`, then `combine_inf_chains()` does paired clustering. But I may be missing something about how this interacts with the disjoint grouping step. Please let me know if this does not match your expectation.

Yes, I think pair cleaning would happen after assembly the way we have it now. Pair cleaning does do a lot of iterating over things, so there might be a big benefit to doing that only on disjoint subsets (as with pair info extraction). But I'll  do some more testing on it as is to think through the different options. As you say I think it does do all the steps now.   

> On paired apportionment (grouping on both heavy+light CDR3 lengths): I think that is a good idea worth exploring, but it would require reading paired info before grouping, which the current step 1 does not do. I would plan this for Phase 2 alongside hfrac sub-grouping rather than trying to fit it into this PR.

Yeah, that part would be more complicated, that's correct.

> **Arg stripping verification**

I'll write something to check this.


> On the naming: you are right that "strip_paired_args" is not fully accurate. Some of the args are not paired-specific but are stripped because their effects are handled at the outer level (like `--dont-calculate-annotations`). I am open to a better name if you have a suggestion.

oh, i wasn't complaining about the name, that's fine, just the comment/docstr I think.

> **input_metafnames**
> 
> You were right to question this. I found that moving the default into the disjoint-only block broke the non-disjoint subset-partition path. `run_subsets()` and `get_cmd()` both use `args.input_metafnames` for pair info handling and forwarding. Fixed in `e1fb18759`, the default is back at the top of `subset_partition()` where both paths can use it.

great

> **get_sw_cache_path**
> 
> The reason it uses glob instead of constructing the path deterministically is that `run_disjoint_group()` does not read the input file. It reads the SW cache directly to get CDR3 lengths for grouping. The SW cache filename includes a UID hash (`sw-cache-{md5(uids)[:10]}.yaml`) computed from the input UIDs by `PartitionDriver`. To construct that filename, you would need to read the input file to get the UIDs, which defeats the purpose (the whole point is to avoid loading millions of sequences when the cached annotations already have what is needed).
>
> Subset-partition does not have this issue because its subprocesses go through `PartitionDriver`, which reads the input and computes the hash internally. An alternative would be to pass `--sw-cachefname` explicitly from the outer level, similar to how `prep_args()` does it at line 906, but that path computation lives inside `run_all_loci()` closures.

oh yeah trying to reconstruct the hash would be crazy, I was imagining using the arg to set it to <path>/sw-cache.yaml. I'm looking at the subset partition results in the test dir, they all use sw-cache.yaml, I'm guessing as a side effect of running everything as paired data. I'll poke at this as well when I test some more things tomorrow.

psathyrella:
I just pushed a few things. I'm still working on some others, but just wanted to note a few things as I go. Let me know if any of my changes don't make sense.

   - I turned on --dont-find-new-alleles since it'll make the auto cache parameterrs ~twice as fast, and we ~don't~ shouldn't care about the germline sets
   - I say shouldn't, because right now we kind of do, since currently it's using those auto cached sw-only parameters for all steps (partitioning each disjoint subset as well as getting annotations at the end). This is bad, we need hmm annotations
   - i added --sw-cachefname so we don't need to mess with hashes (besides the complication of guessing which one, with the old approach the subset procs weren't finding the sw cache at all, so were having to rerun smith waterman). The final annotation processes during paired merging also wasn't finding them
   - I think we want to write git info? Seems good to have that in, I assume that was just from copying from test.py
   - the auto parameter caching still isn't passing most args through, which will do bad things if users are setting args on the command line
   - combining chains on the concatenated output I think kind of avoids the goal of disjoint grouping -- this is for instance merging paired partitions, which e.g. loops over everybody's pair info looking for partners, when the point of disjoint grouping is we know they're disjoint, so should just be concatenating. But we can't because the disjoint stuff is running on each locus individually.


Will update when I push again.

EDIT ok darn, this is hard -- we can't apportion by joint heavy+light cdr3 length since we don't know the actual pairings until pair cleaning happens after partitioning. hmm.

Also, sorry, i rmd the test files for the time being to make diffs and merges more manageable until we get the overall workflow nailed down.

psathyrella:
It turns out I was wrong about being able to combine (what we're calling...) subset partition and disjoint grouping, because apportioning droplets and apportioning by cdr3 length are in practice incompatible. (I say "what we're calling" since I think the current names are confusing because they're both subsetting and partitioning and grouping.) Apportioning droplets requires (of course) putting all sequences from a droplet in the same subset. Now the piece we thought we could add was making those subsets disjoint using cdr3 length. If we always had exactly one cell per droplet, and if each cell always had exactly one heavy and one light chain sequence, this would be easy: simultaneously group by heavy and light cdr3 length, so a group would be, say igh 33+igk 27. But there's lots of experimental and biological processes that break this (see e.g. allelic inclusion + related stuff [here](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1010723)). So taking this example with sequences in four droplets (colors indicate clonal families until I ran out of colors):

p


We try to group together h 33 sequences, so also include k 27; but then k 30 and k27 have to come along for the ride, and then also h 27, and so on. In principle, you could get all sequences required to be in one subset. We have proper pairing information (i.e. exactly 0 or 1 pair per sequence) only after partitioning, since we need partition information to fix the pair info.

This unfortunately seems to mean that we can't implement this the simple way with disjoint grouping being just an alternative apportionment scheme for the existing subset partition method, since it seems that disjoint grouping as a concept only really works for single chain runs. So I think what we're left with is what you were advocating (whoops, sorry): nested subsetting, first apportioning droplets into random subsets, and then within each such subset, using disjoint grouping for the single chain partition runs. The downsides to this are 1) having two separate subsetting schemes is super confusing and 2) the actual paired clustering and pair cleaning algorithms have to run on the full sample rather than on each disjoint subset.

I also found this useful:

p

s-canchi:
**On the architectural concern**

I want to make sure I understand the concern correctly. The current implementation groups sequences by CDR3 length per-locus independently. It does not attempt to group droplets by joint heavy+light CDR3 length. The droplet/pairing info is metadata passed through via `--input-metafnames`, and pair cleaning runs on the full assembled output at the end, same as subset-partition does after merging subsets.

Is the concern specifically about the transitive closure issue with joint heavy+light grouping? Or is there a problem with the per-locus grouping approach itself that I am not seeing?

Could you also clarify the nested subsetting direction, random subsets first then CDR3 grouping within? I am not sure I understand why that order over CDR3 grouping first with subset-partition within large groups if needed.

**On SW-only parameters**

You are right, the current code incorrectly reuses the SW-only parameters from the grouping step for per-group partition. The SW step should only be for the initial fast CDR3 length splitting. Each group should then run full cache-parameters + partition with its own parameter dir as if it were a new dataset. I will fix this unless you already have it in progress.

psathyrella:
> **On the architectural concern**
> 
> I want to make sure I understand the concern correctly. The current implementation groups sequences by CDR3 length per-locus independently. It does not attempt to group droplets by joint heavy+light CDR3 length. The droplet/pairing info is metadata passed through via `--input-metafnames`, and pair cleaning runs on the full assembled output at the end, same as subset-partition does after merging subsets.

Yes, the pair info is passed through and used on the full assembled output to run both the pair cleaning and paired clustering algorithms (i.e. 'merge-paired-partititons'/combine_inf_chains()). I don't know that it's exactly the same as subset partition as each subset partition process is doing full paired cleaning and paired clustering workflow (see e.g. ` test/paired/new-results-slow/subset-partition-new-simu/isub-0/partition.log`). The final merged partition step in subset partition also runs the equivalent of merge-paired-partitions, and of course all the pair info is already cleaned, but the paired clustering algorithm still needs to run again ("synchronizing heavy and light chain cluster paths" in log). 

> Is the concern specifically about the transitive closure issue with joint heavy+light grouping? Or is there a problem with the per-locus grouping approach itself that I am not seeing?

Sorry I don't think I understand this, perhaps better to chat on zoom. But the issue is that it does not seem possible, in general, to apportion both by droplets and cdr3 lengths/disjointly.
 
> Could you also clarify the nested subsetting direction, random subsets first then CDR3 grouping within? I am not sure I understand why that order over CDR3 grouping first with subset-partition within large groups if needed.

It's hard to say "first" vs "second" as it depends what's being run, but the disjoint stuff is single locus, so any single locus partition command where the arg was set will split by cdr3 lengths and run them independently. So if it's set during subset partition, it will group droplets by subsets, and within each subset's single chain runs, disjoint grouping will happen.

> **On SW-only parameters**
> 
> You are right, the current code incorrectly reuses the SW-only parameters from the grouping step for per-group partition. The SW step should only be for the initial fast CDR3 length splitting. Each group should then run full cache-parameters + partition with its own parameter dir as if it were a new dataset. I will fix this unless you already have it in progress.

Oh yeah it's fixed in that ^ commit.

-------------

I'm pretty sure what I committed there ^ is roughly what we want. Please let me know if I a) did anything dumb or b) changed anything that messes with your workflow, especially the slurm stuff. I still need to run through some more todos I have written down, as well as test on a bunch more combinations of arguments, and read through some more log files. I should be able to finish tomorrow though.

I also changed --disjoint-group to --disjoint-groups, and 'disjoint-group' action to 'create-disjoint-groups', in both cases because I kept getting confused whether '-group' was a noun or a verb. But feel free to also disagree with these changes!

s-canchi:
Thanks for the restructure and clarifications. I want to confirm my understanding of the code paths since this is important for Phase 2.

After reading the restructured code, I understand that `--disjoint-groups` is now a partition-level flag that works in two contexts:

1. `partis partition --disjoint-groups`: CDR3 grouping up front, per-group partition, assemble, then `combine_inf_chains()` once on the full output.

2. `partis subset-partition --disjoint-groups`: the flag passes through to each subset's partition call (not stripped from per-subset commands, only from the merged re-partition). So each subset runs the full paired workflow with CDR3-based partition optimization internally, including per-subset pair cleaning.

I had previously misunderstood the level at which disjoint grouping would apply. Making it a partition-level optimization that composes naturally with subset-partition is much cleaner than the previous integration in `subset_partition()`.

For oversized CDR3 groups in the standalone HPC path, the correct fallback is standard subset-partition on that group since all sequences already share the same CDR3 length and further grouping would be meaningless. I have not tested the standalone actions yet (that is planned for Phase 2) but based on the code they should work since they are thin wrappers around the same functions.

Is that correct?

**On the naming changes**

`--disjoint-groups` and `create-disjoint-groups` are clearer, agreed.

**Thinking ahead to Phase 2 (hfrac grouping)**

For oversized CDR3 groups, I am thinking about using hfrac-based bounds to split them into smaller biologically coherent sub-groups instead of falling back to random subset-partition. Two questions on this:

1. Would the right comparison be hfrac-based sub-groups vs random splits (subset-partition) on data that already shares the same CDR3 length, in terms of accuracy and speed?

2. If hfrac sub-grouping can get group sizes small enough, would that open the door to using HA instead of vsearch for those groups? At current group sizes I do not think HA would be triggered.

psathyrella:
> Thanks for the restructure and clarifications. I want to confirm my understanding of the code paths since this is important for Phase 2.
> 
> After reading the restructured code, I understand that `--disjoint-groups` is now a partition-level flag that works in two contexts:
> 
>     1. `partis partition --disjoint-groups`: CDR3 grouping up front, per-group partition, assemble, then `combine_inf_chains()` once on the full output.
> 
>     2. `partis subset-partition --disjoint-groups`: the flag passes through to each subset's partition call (not stripped from per-subset commands, only from the merged re-partition). So each subset runs the full paired workflow with CDR3-based partition optimization internally, including per-subset pair cleaning.

Yes, except I don't know that I'd say they're different contexts, it's just that during the course of subset-partition, single chain partitioning occurs, and during single chain partionining we can use disjoint groups. 

 
> I had previously misunderstood the level at which disjoint grouping would apply. Making it a partition-level optimization that composes naturally with subset-partition is much cleaner than the previous integration in `subset_partition()`.

I don't know that you'd misunderstood -- we hadn't decided yet, since there were a lot of options and it wasn't yet clear which of them made sense. For instance I also thought through caching parameters also in disjoint groups, but a) germline inference loses sensitivity on smaller samples and b) we already had to run at least sw parameters on everything first, so it's a lot cleaner to just run full sw+hmm parameter caching on the whole sample and be done with caching parameters.  
 
> For oversized CDR3 groups in the standalone HPC path, the correct fallback is standard subset-partition on that group since all sequences already share the same CDR3 length and further grouping would be meaningless. I have not tested the standalone actions yet (that is planned for Phase 2) but based on the code they should work since they are thin wrappers around the same functions.
 
I think running subset-partition on a single cdr3 group probably work fine, although you'd have to run it by hand since the current implementation runs plain partition on each disjoint group. I don't think we should think about further splitting of disjoint subsets until we encounter one that's actually inconveniently large (let me know if I'm wrong and you already have). Splitting by cdr3 should reduce the maximum size of anything we're partitioning by more than a factor of 10, so I would imagine going further than that would start being slower from per-process infrastructure overwhelming any potential benefit to more splitting. If we need to split further, I think we should do the hfrac splitting, but I also don't think we should do that until we actually get stuck on a cdr3 subset that's taking annoyingly long.       

> **On the naming changes**
> 
> `--disjoint-groups` and `create-disjoint-groups` are clearer, agreed.
> 
> **Thinking ahead to Phase 2 (hfrac grouping)**
> 
> For oversized CDR3 groups, I am thinking about using hfrac-based bounds to split them into smaller biologically coherent sub-groups instead of falling back to random subset-partition. Two questions on this:
> 
>     1. Would the right comparison be hfrac-based sub-groups vs random splits (subset-partition) on data that already shares the same CDR3 length, in terms of accuracy and speed?
> 
>     2. If hfrac sub-grouping can get group sizes small enough, would that open the door to using HA instead of vsearch for those groups? At current group sizes I do not think HA would be triggered.

I don't think we want to consider further splitting until a sample forces us to, but I think as is we should change the thresholds to use HA. I would just run with and without --disjoint-groups, forcing non-vsearch clustering, on increasingly large samples to compare how long things take. Probably turn on --dont-calculate-annotations since that'd be the same for both, and also use pre cached parameters so they're not remaking those

s-canchi:
One clarification on cache-parameters: in our earlier implementation I was running full cache-parameters per CDR3 group (after SW-only on the full data for grouping). You removed this in the restructure and run full cache-parameters once on the entire dataset instead. Is per-group HMM parameter estimation actually inaccurate (biased deletion/insertion distributions from seeing only one CDR3 length), or is the concern primarily about germline inference sensitivity on smaller samples? The answer matters for scaling since at 4M+ we cannot run cache-parameters on the full dataset in one process.

For soto-bcr scale, I am thinking about chunked cache-parameters: split the input into random chunks (e.g., 500K sequences each), run cache-parameters independently on each chunk as parallel SLURM jobs, merge the HMM parameter dirs (small files, `utils.merge_parameter_dirs()` exists). Random chunking preserves CDR3 length diversity so each chunk sees the full distribution. The SW caches would stay as separate per-chunk files (not merged into one giant file) to avoid I/O stalls. CDR3 grouping would then read across chunk SW caches, extracting only CDR3 length and UID to build global groups, and write per-group SW cache subsets directly from the chunks.

For the HPC enter-exit path at scale this would need two changes: (1) a standalone `merge-parameter-dirs` action exposing `utils.merge_parameter_dirs()` so chunked cache-parameters results can be merged without a parent process, and (2) `create-disjoint-groups` accepting multiple SW cache paths instead of one so it can read across chunks without requiring a merged SW cache file.

Does this seem like the right direction for scaling to soto-bcr, or do you see issues with this approach?

psathyrella:
> One clarification on cache-parameters: in our earlier implementation I was running full cache-parameters per CDR3 group (after SW-only on the full data for grouping). You removed this in the restructure and run full cache-parameters once on the entire dataset instead. Is per-group HMM parameter estimation actually inaccurate (biased deletion/insertion distributions from seeing only one CDR3 length), or is the concern primarily about germline inference sensitivity on smaller samples? The answer matters for scaling since at 4M+ we cannot run cache-parameters on the full dataset in one process.

I'm pretty sure the subprocesses were using the initial sw-only parameter cache? But in either case, yes I would expect many parameters to correlate with cdr3 length, so per-group parameters would be quite different; but for non-germline-set parameters merging them (with proper weighting for sample size, which I don't think is implemented) I think would work fine. But calling new genes requires thresholds, so if splitting apart the sample brings us below these thresholds, we won't call any new genes in the subprocesses, so they won't be there even if we merge the partitions. Running on random subsets has none of these problems, it's just a matter of making sure they're large enough to be representative (and of merging/finding the sw cache files).

Really? I could've sworn I'd cached parameters on at least 4M sequences on thneed, which I think is pretty puny these days. And didn't we cache sw parameters on the whole sample? hmm parameter caching doesn't usually take significantly longer than sw, it just runs afterward. I could be misremembering, though. In any case, this is exactly what subset-partition/subset-annotate do already, isn't it? Split apart, cache parameters, merge parameters; then they partition/annotate, but you could run without that step.

> For soto-bcr scale, I am thinking about chunked cache-parameters: split the input into random chunks (e.g., 500K sequences each), run cache-parameters independently on each chunk as parallel SLURM jobs, merge the HMM parameter dirs (small files, `utils.merge_parameter_dirs()` exists). Random chunking preserves CDR3 length diversity so each chunk sees the full distribution. The SW caches would stay as separate per-chunk files (not merged into one giant file) to avoid I/O stalls. CDR3 grouping would then read across chunk SW caches, extracting only CDR3 length and UID to build global groups, and write per-group SW cache subsets directly from the chunks.

Is the i/o stall really a recurring issue? I've just never see it, and I always have /fh/fast mounted via sshfs, which seems much more fragile and crappy than what's used on site. I mean I definitely could be wrong but my feeling is that our files are just not that big, they should fit in memory easily, so this feels like a lot of complication and optimization that we don't necessarily need. And I thought we decided before to merge all outputs (parameters and partition files), so we'd be merging anyway.  

> For the HPC enter-exit path at scale this would need two changes: (1) a standalone `merge-parameter-dirs` action exposing `utils.merge_parameter_dirs()` so chunked cache-parameters results can be merged without a parent process, and (2) `create-disjoint-groups` accepting multiple SW cache paths instead of one so it can read across chunks without requiring a merged SW cache file.

I'm not necessarily opposed to making the parameter merging a separate command, but I'm also not sure how that'd be different to running subset-partition/subset-annotate on existing subset parameters, it'd just find and merge them.

psathyrella:
I may be remembering wrong, but weren't the file read issues cropping up in an earlier version of the code that was using bespoke file reading with yaml, rather than the existing functions that use json?

s-canchi:
To clarify my earlier statement about cache-parameters at 4M+: I did not mean it cannot be run at 4M. The soto-bcr benchmarks completed successfully at 4M (~136 GB, ~4.3 hours for cache-parameters). It is at 10M+ where single-process cache-parameters becomes infeasible on standard cluster nodes, and at 17M per donor it failed outright (140 GB, OOM-killed).

You are right that I/O stalls are not the issue. Memory for reading soto-bcr benchmark SW cache files with json.load:

| Dataset | File Size | Peak Memory |
|---------|-----------|-------------|
| 1M seqs | 2.3 GB | 11 GB |
| 2M seqs | 4.6 GB | ~22 GB |

The constraint is in-memory representation: 11 GB to hold 1M parsed annotations in Python dicts. At 17M per donor that would be ~190 GB just for the SW cache read, on top of the ~34 GB/M that cache-parameters itself needs (from soto-bcr benchmarks where full cache-parameters on 17M failed at 140 GB and was OOM-killed).

On subset-partition already handling this: the issue is that `run_subsets()` reads the entire input file into memory for the initial subsetting step. At 17M that hits the same memory wall before any subprocess starts.

The chunked approach I described is compatible with what you suggested (split apart, cache parameters, merge parameters). The only difference is the splitting must happen externally (not inside a parent process that loads everything) and the CDR3 grouping step needs to read across chunk SW caches sequentially rather than requiring one merged file. Reading one 1M-sequence chunk SW cache at a time peaks at ~11 GB on the soto-bcr IGH benchmarks, so this should run on standard cluster nodes regardless of total dataset size.

s-canchi:
One clarification on the infrastructure side: looking at this more closely, the only partis-internal change needed for the chunked approach is `create-disjoint-groups` accepting multiple SW cache paths instead of one. The rest (input chunking, parameter dir merging, job orchestration) can live in external scripts using `utils.merge_parameter_dirs()` as a library call. This would avoid any major refactor or code changes to partis. Thoughts?

psathyrella:
Ah, thanks, I get it. I'd forgotten that the samples are much larger than 4M. It definitely makes sense that memory becomes an issue at that size.

I for sure agree it could make sense to implement most of this in your separate code with just the change to `create-disjoint-groups`. I'm just a little worried that you'll be writing all this useful code that gets lost for next time we want to do it. I'm looking at subset-partition, it looks to me like it only reads the full input file to get input_info, which it then immediately converts to seqfos that consist only of uid and sequence, so it seems like what you're describing could be a simple memory optimization here.

>  The only difference is the splitting must happen externally (not inside a parent process that loads everything) and the CDR3 grouping step needs to read across chunk SW caches sequentially rather than requiring one merged file.

How would the external split everything without loading every sequence? Or do you mean like it only loads seq, cdr3 len, uid? Also, this paragraph seems to be talking about both disjoint grouping and subset partition, which are incompatible, which may be why I'm confused. 

Also could you remind me, what are we using Soto for? I can't remember where it's described what we're doing with it/what it's like ([this issue](https://github.com/matsengrp/data-central/issues/52) just says to avoid it).

s-canchi:
To clarify the confusion points:

**On external splitting without loading everything**: The external split for cache-parameters is just splitting a FASTA file (line by line, no annotations needed). The CDR3 grouping step is separate and reads SW cache files after cache-parameters has run. These are two different steps, not conflating disjoint grouping and subset-partition.

**On the memory optimization to `run_subsets()`**: I agree that reading only uid + sequence + paired-uids instead of full annotations would reduce memory significantly (from ~11 GB/M to maybe 1-2 GB/M). This would help mid-sized datasets (1M-5M) that go through subset-partition on standard nodes. It is worth doing independently of the large-scale work.

However, even with this optimization, subset-partition's parent process still runs subsets sequentially and stays alive for the full pipeline. At 17M with 10 subsets, that is 10 sequential runs of cache-parameters + partition taking days on one node with no failure recovery if something crashes at hour 47.

**On the code not being wasteful**: The external orchestration scripts would call the same partis functions and standalone actions (`cache-parameters`, `create-disjoint-groups`, `partition`, `assemble-groups`). The work is in the orchestration layer (splitting, job coordination, parameter dir merging via `utils.merge_parameter_dirs()`), not in reimplementing partis logic. The partis code we developed in this PR is used directly.

**How I envision processing the full soto-bcr dataset (17M per donor):**

1. Split input FASTA into 1M-sequence chunks externally (17 chunks per donor)
2. Run `partis cache-parameters` on each chunk as parallel SLURM jobs (17 jobs, ~34 GB each, ~1 hr each)
3. Merge HMM parameter dirs via `utils.merge_parameter_dirs()` (small script, seconds)
4. Run `partis create-disjoint-groups` reading across chunk SW caches (needs the multiple-path change, ~11 GB peak, minutes)
5. Run `partis partition` per CDR3 group as parallel SLURM jobs (50+ jobs, right-sized per group)
6. Run `partis assemble-groups` per locus (seconds)
7. Run `partis merge-paired-partitions` if paired data

Wall-clock time: days instead of weeks. Memory per job: 34 GB max. Every step is independently restartable. All partis commands are standard except step 4 needing multiple SW cache paths.

**On soto-bcr**: It is a large-scale BCR dataset from Soto et al. with 3 donors, ~56M IGH sequences per donor raw, ~17-20M per donor after filtering for full-length chains (single chain data). The raw data has mixed sequencing methods targeting different regions of the heavy chain, so many sequences are truly 5' truncated (which is why issue #52 flagged it as avoid). After filtering on v_5p_del=0 and j_3p_del=0 we retain the full-length subset. The primary use is for scaling dasm2 to large sample sizes. Details in data-central issue #35.

psathyrella:
> However, even with this optimization, subset-partition's parent process still runs subsets sequentially and stays alive for the full pipeline. At 17M with 10 subsets, that is 10 sequential runs of cache-parameters + partition taking days on one node with no failure recovery if something crashes at hour 47.

Yeah, totally agree, but I'm not suggesting leaving subset partition as-is, I'm just saying that the only change to it to to enable running all steps separately would be adding a sys.exit(0) or something after it writes the input files for the various subsets. I also don't understand no failure recovery -- everything seems to me would be on disk and reread on rerun the same as if run separately. Also note that while I have no objection to adding an exit to facilitate running subset jobs with outside code/by hand, as noted above it's a trivial change to swap simplerun to run_cmds so they're not sequential.  
 
> **On the code not being wasteful**: The external orchestration scripts would call the same partis functions and standalone actions (`cache-parameters`, `create-disjoint-groups`, `partition`, `assemble-groups`). The work is in the orchestration layer (splitting, job coordination, parameter dir merging via `utils.merge_parameter_dirs()`), not in reimplementing partis logic. The partis code we developed in this PR is used directly.

Ah, sorry, I wasn't saying it's wasteful because it made new code, I'm just saying that if the orchestration for super large samples is worth writing for this sample, it might be useful to have in partis so we can use it again. But either way!
 
> **On soto-bcr**: It is a large-scale BCR dataset from Soto et al. with 3 donors, ~56M IGH sequences per donor raw, ~17-20M per donor after filtering for full-length chains (single chain data). The raw data has mixed sequencing methods targeting different regions of the heavy chain, so many sequences are truly 5' truncated (which is why issue #52 flagged it as avoid). After filtering on v_5p_del=0 and j_3p_del=0 we retain the full-length subset. The primary use is for scaling dasm2 to large sample sizes. Details in data-central issue #35.

Great, thank you for the pointer! 

I think at this point just go ahead and do what makes sense to you, you're much more familiar with the details of what you need to do.

s-canchi:
Thanks for the clarifications and the go-ahead. The subset-partition fixes (memory optimization in `run_subsets()`, `simplerun` -> `run_cmds` for concurrency, early exit after writing subset files) make sense as a separate PR since they benefit all datasets.

For this PR, the remaining items on my side are:
- Update partis docs for renamed flags (`--disjoint-groups`, `create-disjoint-groups`, `--n-max-disjoint-jobs`, `--n-sub-procs`)
- Add support for multiple SW cache paths in `create-disjoint-groups`, since that is core disjoint grouping functionality needed for the chunked cache-parameters approach at scale. The change would be in `create_cdr3_groups()` to accept a list of SW cache files and read them sequentially.

I am waiting on your testing of argument combinations and rebuilt ref results. Should I go ahead and implement the multi-path change now, or wait until your testing is done to avoid conflicts?

psathyrella:
oh, sorry, didn't realize you were waiting for me, go ahead. I have the arg tests written down still but I can do it later. I was not planning to rebuild the ref results now since as currently set up --disjoint-groups is just added to the subset-partition test, and the ref results for all existing tests are updated on dev+main, so we can just update them after we merge, if we need to. Currently the purity/completeness change a tiny bit, but that's not really important since the main validation test is on samples with significant size.

s-canchi:
The docs are updated for the new flags and standalone actions. I also added support for multiple SW cache paths in `create-disjoint-groups` for the chunked cache-parameters workflow at scale. Both changes are pushed.

I think this PR is ready for final review and merge from my side. Following this merge, I plan to:
1. Implement the subset-partition speedups (memory optimization, concurrent subsets, early exit) as a separate PR
2. Run the current disjoint grouping implementation on the soto-bcr benchmark subsets (1M, 2M, 4M) and compare output similarity and runtime against existing standard partition and subset-partition results
3. Based on the CDR3 group size distributions from those runs, evaluate whether hfrac sub-grouping is needed for oversized groups
4. Benchmark HA clustering by re-running partition on the disjoint groups with `--no-naive-vsearch` using the same pre-cached parameters, to determine if group sizes make HA practical

Does this sound reasonable as a plan for Phase 2?

s-canchi:
@psathyrella - Let me know when this might be ready for merge!

psathyrella (approved):

psathyrella:
Yes, sounds great! just one thing below:

>     4. Benchmark HA clustering by re-running partition on the disjoint groups with `--no-naive-vsearch` using the same pre-cached parameters, to determine if group sizes make HA practical

This sounds good, but will potentially be inconclusive -- HA could be annoyingly slow on by far the largest sample we've ever run on, but we might still want to crank up the threshold for turning on --fast. I could look into this, but since it sounds like you have all the infrastructure in place already, perhaps you could also run some increasingly-large (or decreasingly-small, whatevs) subsets to get us some info on what we might want to set this threshold to.

PR #346, opened 2026-03-31
subset-partition: memory optimization, concurrent subsets, early exit

s-canchi:
#### Summary

- Strip `input_info` to meta fields after subsetting to reduce memory in `run_subsets()` (estimated 5-10x reduction at scale)
- Switch `utils.simplerun()` to `utils.run_cmds()` for concurrent subset execution (14% speedup on test data)
- Add `--write-subsets-only` flag to write per-subset input files and exit, enabling external job orchestration

#### Test plan

- `partis-test.py --quick` passed
- `partis-test.py --paired --no-simu` passed
- `partis-test.py --slow --paired` passed

All tests pass. The subset-partition test includes `--disjoint-groups` so the composed path is validated.


psathyrella (approved):
Looks good!

psathyrella:
@s-canchi is it not possible to use the same arg to limit the number of concurrent subprocesses for disjoint and subset? It seems like it'll be hard to keep them straight. Maybe --n-max-subprocs or something?

s-canchi:
Good point. Should I open a PR to fix the variable naming?

psathyrella:
That would be fine, but you can also just add the change to whatever other changes I think you're working on now.

PR #347, opened 2026-04-01
unify --n-max-disjoint-jobs and --n-max-subset-jobs into --n-max-subprocs

s-canchi:
Per feedback on PR #346: unify the two separate concurrent job limit flags into a single `--n-max-subprocs` arg (default 2). Defined as a partition arg so subset-partition inherits it via `sub_arg_groups`.

#### Test plan

- `partis-test.py --paired --no-simu` passed


psathyrella (approved):
looks good!

PR #348, opened 2026-04-13
fix standalone disjoint grouping actions for batch workflows

s-canchi:
#### Summary

Fixes found while validating the disjoint grouping pipeline on soto-bcr benchmark subsets (1M sequences). These address compatibility issues when running create-disjoint-groups, partition, and assemble-groups as standalone batch jobs.

- Use recursive glob for SW cache discovery in nested directory structures
- Skip workdir-exists check for create-disjoint-groups and assemble-groups (these actions create subdirectories within an existing workdir)
- Skip groups with no partition output in assembly instead of erroring (supports excluding tiny groups from partition)
- Account for skipped groups in assembly sequence count validation
- Auto-discover partition files in group dirs when manifest partition_path is not set (supports standalone partition jobs that do not update the manifest)
- Document standalone workflow notes for unpaired data and at-scale parameter merging

#### Validation

Tested on soto-bcr Donor03 1M subset using the standalone batch workflow (chunked cache-parameters, multi-cache CDR3 grouping, per-group vsearch partition, assembly). Assembled output compared against v0 standard partition:

- 91% cluster consistency
- 28,257 vs 27,676 clusters (filtered to size >3)
- 574,020 shared sequences (99.5% overlap)
- Largest cluster identical (5,333 in both)

Per-step wall time comparison:

| Step | Disjoint vsearch (parallel) | Standard partition (v0) |
|------|----------------------------|------------------------|
| Cache-params | 42 min (2 chunks) | ~1 hr |
| CDR3 grouping | 5 min | n/a |
| Partition | 9 min (longest group) | 2:30 |
| Assembly | 58s | n/a |
| Total | ~60 min | ~3.5 hrs |

Max per-job memory: 17.5 GB (cache-params chunk) vs 20.7 GB (standard partition).

### DIFF
--- a/docs/subcommands.md
+++ b/docs/subcommands.md
@@ -165,12 +165,15 @@ Since normal annotation, unlike partitioning, is easily split up into independen
 Split sequences for a single locus into disjoint CDR3 length groups, writing per-group FASTAs, per-group SW cache subsets, and a manifest to the output directory.
 Requires `--locus` and `--parameter-dir` (which must already contain an SW cache from a prior `cache-parameters` run).
 `--sw-cachefname` can point to a single SW cache file, a colon-separated list of files, or a directory containing multiple SW caches (e.g. from running cache-parameters independently on each part of a split input).
+When running `cache-parameters` independently on unpaired data for this purpose, pass `--paired-loci --no-pairing-info` so the parameter directory layout is compatible with downstream functions.
+At scale, merge only HMM parameters and germline sets across parts externally, and pass the directory of per-part SW caches to `--sw-cachefname` rather than merging them into one file.
 This is the standalone version of the grouping step in `--disjoint-groups` (see [above](#disjoint-groups)), intended for workflows where each step is submitted as a separate batch job.
 
 ### assemble-groups
 
 Concatenate per-group partition results from disjoint grouping into a single output file for one locus.
 Requires `--locus` and `--outfname`.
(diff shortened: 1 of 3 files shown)


psathyrella:
Looks good! just a couple comments:
  - in disjointgrouper, I don't understand how we'd end up with multiple partition files with different subdirs? Can we not just enforce a particular subdir pattern?
  - same comment for sw-cache. It just seems dangerous to glob recursively unless we really need to

s-canchi:
Good points on both.

For partition auto-discovery: replaced the recursive glob with a check for the specific expected path (`groups/cdr3-N/partition-{locus}.yaml`). This is the path both the integrated and standalone paths write to. The recursive glob was a workaround for the workdir nesting issue (which is also fixed on this branch), so it is no longer needed.

For SW cache resolution: replaced the recursive glob with explicit expected patterns tried in order:
1. `*/parameters/*/sw-cache*.yaml` (chunked paired layout)
2. `*/parameters/sw-cache*.yaml` (chunked flat unpaired layout)
3. `*/sw-cache*.yaml` (single level)
4. `sw-cache*.yaml` (same directory)

This covers paired, unpaired, and chunked cases without risking stale file discovery from deeply nested directories. The error message lists which patterns were checked if none match.

psathyrella (commented):
looks good, thanks!

PR #349, opened 2026-04-19
hfrac sub-grouping with TCM round 2

s-canchi:
Adds `--hfrac` flag for naive hamming fraction sub-grouping within CDR3 groups. Uses a two-round approach: round 1 vsearch greedy clustering at `hi_bound`, round 2 transitive closure merge (TCM) on centroids at a configurable multiplier of `hi_bound`.

Validation results and discussion on #337.

#### Changes

- `partis/disjointgrouper.py`: TCM round 2 (`_build_round2_tcm_cmd`, `_tcm_merge_from_pairs`), bin-packing cap, two-pass hfrac for multi-cache path, assembly validation fix for filtered partitions
- `bin/partis`: `--hfrac`, `--hfrac-merge-factor` (default 3.0), `--hfrac-max-bin-size` (default 25K) as hidden dev flags. Parameter dir check for missing loci.
- `partis/processargs.py`: `--hfrac` requires `--disjoint-groups` validation


s-canchi:
#### Planned validation runs (no code changes)

These use the merged code and external scripts to validate behavior at scale and on paired data. None are expected to result in code changes.

- [x] Hybrid vsearch+HA correction on 1M soto-bcr subset: confirm simulated data trends hold on real data. Existing vsearch and HA baselines (with and without filter) are available for comparison.
- [x]  Hybrid at 10% SHM across seeds 1 and 2: confirm multi-seed consistency beyond 5% SHM.
- [x] One Wyatt 10x paired donor (~365K sequences): compare vsearch+paired vs HA+paired clustering to determine whether paired clustering offsets vsearch over-merging.

partis/disjointgrouper.py
```diff
@@ -70,11 +227,283 @@ def write_group_sw_caches(groups, glfo, annotation_list, outdir, locus):
         utils.write_annotations(sw_cache_path, glfo, antns_by_c3len.get(c3len, []), utils.sw_cache_headers)
 
 # ----------------------------------------------------------------------------------------
-def write_manifest(group_infos, outdir, locus, total_input, n_failed, parameter_dir=None):
+def _apply_hfrac_and_write(groups, hi_bound, outdir, locus, glfo, annotation_list, merge_factor=3.0, max_bin_size=100000):
```
psathyrella: This function seems awfully large, would it make sense to split it up?

partis/disjointgrouper.py
```diff
@@ -70,11 +227,283 @@ def write_group_sw_caches(groups, glfo, annotation_list, outdir, locus):
...
+# ----------------------------------------------------------------------------------------
+def _apply_hfrac_two_pass(groups, hi_bound, outdir, locus, glfo, merge_factor=3.0, max_bin_size=100000):
+    # memory-efficient hfrac for multi-cache path: reads one CDR3 group SW cache at a time
+    # round 1 pass 1: write naive FASTAs, build vsearch commands
+    # round 1 dispatch: run all vsearch in parallel
+    # round 2 (if merge_factor > 0): TCM on round 1 centroids via --allpairs_global
+    # pass 2: re-read SW caches, parse results, write sub-group outputs
+    import shutil
+
+    min_group_size = 100
```
psathyrella: Do we need this hard coded within the function? I would think at least having it in the function signature (with description) would be clearer.

partis/disjointgrouper.py
```diff
@@ -70,11 +227,283 @@ def write_group_sw_caches(groups, glfo, annotation_list, outdir, locus):
...
+    vsearch_binary = '%s/bin/vsearch-2.4.3-%s-x86_64' % (utils.get_partis_dir(), utils.get_platform_binstr())
```
psathyrella: Can we avoid hardcoding version? Or at least combine all the places we need the vsearch binary to use one function that finds it in utils.py?

partis/disjointgrouper.py
```diff
@@ -70,11 +227,283 @@ def write_group_sw_caches(groups, glfo, annotation_list, outdir, locus):
...
+                'fasta_path' : rel_fasta,
+                'partition_path' : None,
+            })
+            flattened_groups[(c3len, isub)] = sub_seqfos
+        if len(sub_groups_list) > 1:
+            print('        cdr3-%d: %d seqs -> %d sub-groups (sizes: %s)' % (c3len, len(seqfos), len(sub_groups_list), ' '.join(str(len(sg)) for sg in sub_groups_list)))
+    return flattened_groups, all_group_infos
+
+# ----------------------------------------------------------------------------------------
+def _apply_hfrac_two_pass(groups, hi_bound, outdir, locus, glfo, merge_factor=3.0, max_bin_size=100000):
```
psathyrella: _apply_hfrac_and_write( and _apply_hfrac_two_pass( seem both really long and really similar -- i don't fully understand what they're doing, but are they really different enough that they need to be separate?

partis/disjointgrouper.py
s-canchi: Good point. Created a helper fn and moved it to utils.py.

partis/disjointgrouper.py
s-canchi: Removed the hard coded version and moved to fn signature instead.

partis/disjointgrouper.py
s-canchi: Good point. I split the core implementation into three helper fns reducing the scope of this fn drastically.

partis/disjointgrouper.py
s-canchi: The two functions are similar because they share the same logic but differ in how they read input. This is now refactored to reuse core logic helpers instead of duplication across two fns.

psathyrella:
Looks good! thanks

s-canchi:
#### Wyatt 10x paired results (final checklist item)

Ran on one Wyatt 10x paired donor (~365K sequences, fs-d1). Three configurations compared:

| Locus pair | Metric | Baseline (vsearch+paired) | Disjoint+hfrac (vsearch) | HA+disjoint+hfrac |
|---|---|---|---|---|
| igh+igk | clusters | 153,263 | 153,169 | 147,946 |
| igh+igk | singletons | 132,975 | 132,939 | 127,966 |
| igh+igk | sequences | 226,017 | 226,018 | 225,994 |
| igh+igk | max cluster | 455 | 482 | 949 |
| igh+igl | clusters | 97,143 | 97,385 | 93,203 |
| igh+igl | singletons | 84,111 | 84,347 | 80,085 |
| igh+igl | sequences | 138,686 | 138,488 | 138,612 |
| igh+igl | max cluster | 219 | 202 | 398 |

Vsearch disjoint+hfrac is essentially identical to baseline, confirming no degradation from the hfrac code path on real paired data. HA finds fewer, larger clusters as expected from its more aggressive merging.

The original checklist question was whether hybrid adds enough value on top of paired clustering to justify the compute. Paired clustering already significantly helps with vsearch light chain over-merging. Whether hybrid provides additional benefit in the paired context is not yet answered and will be addressed in the follow-up hybrid pipeline PR.

This PR is considered complete. The post-refinement pipeline (hybrid + three-step integration, paired clustering optimization) will follow in a separate PR on top of the merged hfrac code.

s-canchi:
#### Investigation: hfrac vs baseline partition differences

Per the author's request to poke around the small differences between baseline and disjoint+hfrac. Investigated on 30K simulated data (seed-0, both SHM rates) with ground truth. We expect similar behavior on Wyatt where we observed comparable small differences.

**IGH (well-partitioned regime):**

| SHM | Changed UIDs | Affected clusters (base/hfrac) | Purity (base/hfrac) | Completeness (base/hfrac) |
|---|---|---|---|---|
| 5% | 1,180 / 29,887 (3.9%) | 252 / 268 | 0.80 / 0.85 | 0.92 / 0.93 |
| 10% | 5,373 / 29,887 (18.0%) | 969 / 1,020 | 0.71 / 0.73 | 0.96 / 0.95 |

Changes are spread evenly across CDR3 lengths with no concentration at hfrac group boundaries. Example clusters show typical HA non-determinism: clusters differ by 1-2 sequences, sometimes splitting an impure cluster into purer pieces, sometimes the reverse. Hfrac is marginally better on the affected clusters.

**IGK (over-merged regime, ~4% purity):**

| SHM | Changed UIDs | Affected clusters (base/hfrac) |
|---|---|---|
| 5% | 10,764 / 29,887 (36.0%) | 54 / 50 |
| 10% | 17,371 / 29,887 (58.1%) | 81 / 87 |

Both partitions are equally over-merged (~4% purity, ~100% completeness). The large number of changed UIDs is sequences shuffling between equally bad mega-clusters (size 200-1,800). Not meaningful differences.

**Conclusion:** The differences come from vsearch non-determinism (different merge/shuffle order when the same sequences are partitioned into different subsets by hfrac grouping). No systematic pattern tied to hfrac boundaries. No edge case to worry about.

psathyrella:
>The original checklist question was whether hybrid adds enough value on top of paired clustering to justify the compute. Paired clustering already significantly helps with vsearch light chain over-merging. Whether hybrid provides additional benefit in the paired context is not yet answered and will be addressed in the follow-up hybrid pipeline PR.

I just wanted to note that it's quite common to have both paired and bulk sequencing for the same subject (the paired data gives you the ability to actually synthesize Abs, but the much deeper bulk gives you more data overall). It seems to me like this would be a situation where you'd probably want hybrid even though you're running with --paired-loci set. Alternatively, I could say that --paired-loci isn't always a great proxy for whether you've got pair info for most of your data.

CC review in next comment. I don't understand the code well enough to be sure that the algorithmic concerns are correct, so use your judgement there. The code quality things seam reasonable. Re: testing, I think we just want --disjoint-groups to be the default when we're done with this, do you think there's any reason not to do that? In which case I think we don't need to add tests, we'd just change the default and bust the test caches before merging to main.

psathyrella:
## Review summary

### What the PR does

Adds `--hfrac` for naive-hamming sub-grouping inside CDR3 groups, on top of `--disjoint-groups`. Two-round algorithm:

- **Round 1** — `vsearch --cluster_fast` at identity `1 - hi_bound` (greedy centroid clustering); `hi_bound` from `utils.get_naive_hamming_bounds(...)`.
- **Round 2 (default on)** — `vsearch --allpairs_global` between round-1 centroids at `1 - min(merge_factor*hi_bound, 0.49)`; pairs build an adjacency graph; BFS finds components; round-1 sub-groups in the same component are merged (TCM).
- **Bin-packing** — if a TCM component exceeds `max_bin_size * 1.2`, First-Fit-Decreasing splits it back into bins (round-1 sub-groups treated as indivisible items).

Each sub-group gets its own FASTA + SW cache at `groups/cdr3-N/sub-groups/sub-NNN/`, is partitioned independently, then concatenated. New flags: `--hfrac`, `--hfrac-merge-factor` (3.0, hidden), `--hfrac-max-bin-size` (100k). Also: a short-circuit in `paircluster.merge_chains.resolve_discordant_clusters` that splits cross-sub-group clusters without computing hfrac, light-chain UID remap after `translate_paired_uids`, a `uid_to_clust_idx` map for O(cluster-size) overlap lookup, multi-chunk glfo recovery via per-CDR3 SW-cache re-read, `validate_assembly` relaxed `!=`→`>` to allow partition-time filtering, and skipping loci with missing parameter dirs.

### Things that look correct

- vsearch UC parsing — S/H rows handled correctly, centroid appears once in `members`.
- TCM connected-components BFS — singletons included, no double-visits.
- Light-chain remap direction matches `translate_paired_uids` (verified against `utils.translate_uids`).
- `c1[0]/c2[0]` representative-uid short-circuit is sound — all cluster members come from one independently-partitioned sub-group.
- `validate_assembly` relaxation is reasonable (still rejects count inflation).
- `dg_uid_groups` plumbing is safe when `--hfrac` is off — the short-circuit only fires within a single CDR3 length, where sub-groups only differ when hfrac is on.
- `--hfrac*` flags correctly stripped from per-group subprocess invocations.

### Issues worth addressing

**Algorithmic**

1. **Bin-packing partially undoes TCM.** `_merge_r1_subgroups_by_components` splits oversized components *after* round 2 has merged them — but TCM exists precisely because round-1 may straddle clonal-family boundaries (the code itself flags this at high SHM). For default `merge_factor=3.0` and `hi_bound≈0.08`, the round-2 threshold is aggressive (24% naive-hfrac), making this plausible. Consider keeping TCM-linked sub-groups together at the cost of a larger bin, or at least warning when bin-packing splits a TCM component.
2. **Silent threshold clamp.** `min(merge_factor * hi_bound, 0.49)` saturates whenever `hi_bound > ~0.163`, after which `merge_factor` has no effect. No warning issued. At minimum, log when the clamp engages.
3. **Single oversized round-1 sub-group emitted whole, no warning.** When one indivisible round-1 sub-group exceeds `cap`, it's emitted (correct for family safety) but silently — partition runtime can blow up later with no signal.
4. **Silent fallback** in `_merge_r1_subgroups_by_components`: `centroid_to_comp.get(centroid, 0)` defaults missing centroids to component 0, which can silently merge unrelated sub-groups. Should be an assertion.

**Code quality / DRY**

5. **`_apply_hfrac_and_write` and `_apply_hfrac_two_pass` are near-duplicates.** Round-1, round-2, and per-c3len bin construction are identical; only the source of `uid_to_antn`/`group_glfo` differs. Extract a shared helper. Names are also misleading — "two-pass" refers to memory strategy, not the two clustering rounds. Suggest `_apply_hfrac_single_cache` / `_apply_hfrac_multi_cache`.
6. **`subgroup_by_naive_hamming` is dead code** — defined but called nowhere, superseded by the round-1/round-2 helpers.
7. **Defaults defined in three places** (`bin/partis` partition subargs, `bin/partis` create-disjoint-groups subargs, `create_cdr3_groups` signature) for both `merge_factor=3.0` and `max_bin_size=100000`. Centralize as module-level constants.
8. **Magic numbers**: `tolerance = 1.2`, `n_procs = min(8, ...)` (duplicated), `0.49` clamp — promote to named constants and document.
9. **Inline imports** of `shutil` and `csv` should be at module level.
10. **Naming**: `uid_sub_groups` (in `paircluster.py`) is actually a uid→group_id map; `uid_to_dg_group` reads better.

**Integration / process**

11. **No tests added** for `--disjoint-groups` or `--hfrac` — regression risk; even a small smoke test would help.
12. **`dg_uid_groups` only consulted on the list-chain side** in `resolve_discordant_clusters`, although built for both loci. May be intentional (consistent with the "split chain decides" pattern), but the call-site comment should make this explicit.

**Non-issues** (verified OK): manifest path generalization, multi-chunk glfo re-read, `--hfrac` stripping from subprocesses, `processargs` validation gating, the `dg_uid_groups` no-hfrac case.

### Bottom line

Algorithm is reasonable and the integration with `merge_chains` and the multi-chunk path is careful. No correctness-blocking bugs found, but the **bin-packing-undoes-TCM** behavior and the **silent threshold clamp** are worth a closer look from a phylogenetic-correctness standpoint, and the **DRY duplication between the two `_apply_hfrac_*` functions plus dead `subgroup_by_naive_hamming`** are worth cleaning up before merge. Adding even a minimal `--hfrac` test would reduce regression risk.

s-canchi:
#### Reply to PR #349 review summary

Thank you for the detailed pass. Replies organized to match your three sections. Code-side fixes for items 1-10 and 12 are in commit `59ef066d8`.

#### Algorithmic (items 1-4)

1. Bin-packing partially undoes TCM. Theoretically valid, but I do not see it engage in any of the runs validated so far. Across 21 hfrac runs (single-chain IGH/IGK/IGL on 30K simulated, 5%, 10%, and 30% SHM):

| dataset | locus | max sub-group size | bin-packing tolerance (1.2 * 100K = 120K) |
|---|---|---|---|
| 5% SHM, 30K | igh | 830 | not engaged |
| 5% SHM, 30K | igk | 11,861 | not engaged |
| 10% SHM, 30K | igh | 1,319 | not engaged |
| 10% SHM, 30K | igk | 11,861 | not engaged |
| 30% SHM, 30K | igh | 3,033 | not engaged |
| 30% SHM, 30K | igk | 18,458 | not engaged |

The post-TCM bin reaches at most ~18.5K (IGK at 30% SHM, an outlier even for hfrac), well under the 100K cap and its 120K tolerance. The hfrac vs baseline diff investigation already posted on this PR also showed that the small differences are vsearch greedy non-determinism, not boundary effects, with no concentration at hfrac group edges.

Bin-packing is therefore a memory-safety net for runs I have not validated yet (Wyatt 10x at 365K showed similar behavior; soto-bcr 4M is the next target). When that target run lands and bin-packing actually fires, I will check whether splitting a TCM component into multiple bins changes the partition vs running it as a single oversized bin, and revisit if it matters. Added a preemptive log line that announces "bin-packing engaged: TCM component of N seqs split into bins".

2. Silent threshold clamp at 0.49. Fires whenever `hi_bound > ~0.163`, which corresponds to roughly >15% SHM at default `merge_factor=3.0`. This is real for the 30% SHM runs above. Added a yellow warning when `merge_factor * hi_bound > MAX_TCM_THRESHOLD` so the user knows `merge_factor` had no effect.

3. Single oversized round-1 sub-group emitted whole, no warning. Same data as item 1: the largest indivisible round-1 sub-group is far below the cap. This case requires a CDR3 group with a clonal family that vsearch greedy clustering does not split. Added a yellow warning when this happens, alongside the bin-packing log line.

4. Silent fallback in `_merge_r1_subgroups_by_components`. Agreed, `centroid_to_comp.get(centroid, 0)` is wrong as defensive code. Every round-1 centroid is fed into round-2 TCM, so a missing centroid is a bug, not a recoverable state. Switched to an assertion.

#### Code quality (items 5-10)

5. Duplication between `_apply_hfrac_*`. Already addressed in commit `d58eab99f`. Shared logic lives in `_build_round1_vsearch_cmds`, `_run_round2_tcm`, and `_write_subgroup_outputs`. The two top-level functions differ only in how they source `uid_to_antn` and `glfo`.

6. Dead `subgroup_by_naive_hamming`. Removed.

7. Defaults in three places. Promoted `merge_factor=3.0`, `max_bin_size=100000`, and `min_group_size=100` to module-level constants in `disjointgrouper.py` and referenced them from the argparse defaults in `bin/partis`.

8. Magic numbers. Named `BIN_PACK_TOLERANCE = 1.2`, `MAX_VSEARCH_PROCS = 8`, and `MAX_TCM_THRESHOLD = 0.49`.

9. Inline imports. Lifted `shutil` and `csv` to module scope.

10. `uid_sub_groups` -> `uid_to_dg_group`. Renamed in `paircluster.py`.

#### Integration (items 11-12)

11. No tests. Agree in principle on flipping `--disjoint-groups` to default in lieu of adding an `--hfrac` test, but I would prefer to defer that until after the soto-bcr 4M validation and after the post-hybrid refinement pipeline lands. Wyatt 365K and 30K simulated both show no degradation, but the regime where disjoint+hfrac genuinely needs to be default is multi-million-sequence, which I want to validate first. Designing the test suite once across disjoint + hfrac + refinement also avoids reworking it. Treating this as a separate PR keeps the test-cache bust scoped.

12. `dg_uid_groups` only consulted on the list-chain side. Intentional, mirroring the "split chain decides" pattern in pair cleaning: the list chain is the one that iterates and makes split/merge decisions, so it is the only one that needs sub-group membership for the hamming-skip optimization. Added a one-line comment at the call site to make this explicit.

s-canchi:
> I just wanted to note that it's quite common to have both paired and bulk sequencing for the same subject (the paired data gives you the ability to actually synthesize Abs, but the much deeper bulk gives you more data overall). It seems to me like this would be a situation where you'd probably want hybrid even though you're running with --paired-loci set. Alternatively, I could say that --paired-loci isn't always a great proxy for whether you've got pair info for most of your data.


This is really important point and I planned to test both. I was just curious to see how much more benefit we get out of the hybrid in a paired context. I imagine the signals are all additive in resolving the collision regime.

psathyrella:
>No tests. Agree in principle on flipping --disjoint-groups to default in lieu of adding an --hfrac test, but I would prefer to defer that until after the soto-bcr 4M validation and after the post-hybrid refinement pipeline lands.

yes, that sounds great. No hurry at all.

Everything else looks great, thanks! Merge whenever you're ready.

PR #354, opened 2026-04-30
zig: trellis hot-path pointer cleanup (items 9, 13, 19)

matsen:
## Summary

Three items from #342 bundled because they all touch the same trellis hot-path concern (reduce indirection per `(position, state, prev_state)` work) and share one validation gate. **The 30k wall-time win predicted by item 9 did not materialize in measurement** — see results below.

### Item 9 — flatten `Model.states` pointer-of-pointers to flat slice

`Model.states: ArrayListUnmanaged(*State)` → `ArrayListUnmanaged(State)`. `stateByIndex(i)` was a slice index plus a heap-pointer dereference; it's now a single `&self.states.items[i]`. Same theme as #352 did for transitions, one level up.

The non-trivial part is keeping `*State` pointers in `states_by_name` and `Transition.to_state_ptr` stable. Solution: count non-init states from the parsed YAML up front, `ensureTotalCapacityPrecise(allocator, n_non_init)`, then `addOneAssumeCapacity` per state. The slice never grows past its initial capacity, so `&states.items[i]` stays valid for the Model's lifetime. A `std.debug.assert(items.len == capacity)` after the parse loop makes the invariant machine-checkable so a future stray `states.append` trips a debug build immediately.

`init` and `ending` stay separately heap-allocated (they're not regular indexed states; specials avoiding hot-path branches). `deinit` reorganized: iterate `states.items` as values, free the slice's backing storage, then handle `init`/`ending` separately.

### Items 13 + 19 — inline trellis helpers, single-seq emission fast path

- `inline` keyword on `cacheViterbiVals`, `cacheForwardVals` (`trellis.zig`) and `emissionLogprobSeqs` (`state.zig`). Under ReleaseFast the compiler likely already inlined these; the keyword removes any doubt and signals "hot-path leaf" to future readers.
- Single-sequence fast path in `emissionLogprobSeqs`: bypasses the one-trip loop and `addWithMinusInfinities(0.0, x)` call when `nSeqs() == 1`. Value-identical for every input emissionLogprob can produce. Fires in annotate-mode single-cluster runs; not in partition's multi-seq cluster joins.

## Validation

| Rung | Result |
|---|---|
| 0 — `zig build test` | pass |
| 1 — `partis-test.py --quick --zig` | pass |
| 2a — `partis-test.py --no-simu --zig` | pass; 0/37 naive_seq diffs, 0/14 logprob diffs vs ref |
| 2b — `partis-test.py --paired --no-simu --zig` | pass (all steps green, including subset-partition with `--max-ccf-fail-frac 0.25`) |
| 3 — 5k paired clean | byte-identical to `/tmp/zig-5k-baseline` (3797 igh + 3802 igk events). Wall 5:05 vs 5:10 baseline (within noise). RSS 467 MB unchanged. |
| 4 — 30k paired clean | byte-identical to `/tmp/zig-30k-baseline` (10802 igh + 10861 igk events). **Wall 1:16:33 vs 1:15:35 topic-tip after #353 — within single-run noise but no measurable win.** RSS 3.04 GB unchanged. |

## Honest assessment of the perf result

Item 9 was predicted by the issue body to be a "Medium impact" win on top of #352. At 30k it didn't show. Likely reasons:

- After #352 already flattened the hottest data access (`State.log_probs[]`), the residual cost of `stateByIndex` pointer-chasing is small relative to other inner-loop work (emission lookup, `addInLogSpace`).
- Heap-allocated States likely sat near-sequentially in memory after `c_allocator` returned consecutive `mmap`'d pages, so the "pointer chase" was rarely a true cache miss.
- The compiler may have been inlining `stateByIndex` such that the indirection was already CSE'd in the inner loop.

This is exactly the scenario the handoff comment warns about: *"the predictions can be wrong (items 1 and 7 are the examples). Always measure."*

**Why ship anyway.** Code quality wins independent of wall-time:
- Replaces ~500 small allocations per `Model.parse()` (one per State) with a single slice allocation. Less alloc/free churn at startup; less work for the allocator's metadata structures.
- Simpler `deinit` (one loop over the flat slice, no per-State `destroy`).
- Eliminates a class of latent ownership bugs (the prior code required `states_by_name` to own `*State` pointers heap-allocated separately from the `states` array — easy to get wrong on extension). The capacity-reservation invariant replaces that with a single explicit assertion.
- Sets up future cache-locality work (item 21, etc.) that would want to iterate the flat State array directly, possibly with SIMD-friendly layouts.

**Tradeoff to acknowledge.** The new `parse()` loop has two structurally different branches (init heap-allocates and heap-destroys; non-init slot-writes into a pre-reserved slice with an order-dependent `errdefer` that must `deinit` before `pop`). That is a real complexity increase vs the old uniform "every state is heap-allocated the same way" pattern. The deinit improvement is genuine; the parse improvement is not — it's a tradeoff favoring lifetime simplicity (one slice owns everything) over construction-time symmetry. The asymmetric cleanup is documented at the call site; whether the tradeoff is worth it depends on how much weight you put on the resulting deinit clarity vs the parse-loop complexity.

If reviewers prefer to drop item 9 on the no-measurable-win basis, the commit is independent and easy to revert.

## Test plan

- [x] Rungs 0–3 (paired iteration gate) all green
- [x] 30k byte-equality vs `/tmp/zig-30k-baseline`
- [x] zig-code-reviewer pass; comments tightened on errdefer ordering invariant and single-seq fast-path identity argument
- [x] clean-code-reviewer pass; fixed STATE_MAX off-by-one, added capacity-invariant assertion, merged duplicate iteration over `data.states`, tightened "bit-identical" wording

### DIFF
--- a/packages/zig-core/src/ham/model.zig
+++ b/packages/zig-core/src/ham/model.zig
@@ -131,18 +131,22 @@ pub const Model = struct {
         }
         for (td.symbols.items) |sym| try trk.addSymbol(allocator, sym);
 
-        // Collect state names for transition validation
+        // Collect state names for transition validation, and count non-init
+        // states so we can reserve flat-slice capacity in one shot. The capacity
+        // reservation is what makes `&states.items[i]` pointers stable — without
+        // it, addOneAssumeCapacity below would assert.
         var state_names: std.ArrayListUnmanaged([]const u8) = .{};
         defer state_names.deinit(allocator);
-        for (data.states.items) |*sd| try state_names.append(allocator, sd.name);
-
-        // Reserve flat-slice capacity for non-init states up front so addOne
-        // never reallocates — that's what makes `&states.items[i]` stable.
         var n_non_init: usize = 0;
         for (data.states.items) |*sd| {
+            try state_names.append(allocator, sd.name);
             if (!std.mem.eql(u8, sd.name, "init")) n_non_init += 1;
         }
-        if (n_non_init >= @import("state.zig").STATE_MAX) return error.TooManyStates;
+        // Match the C++/old behavior of allowing exactly STATE_MAX non-init
+        // states. The old code checked `items.len >= STATE_MAX` immediately
+        // before each append, so the (N+1)-th append (when items.len was
+        // already N==STATE_MAX) was the one that errored.
+        if (n_non_init > @import("state.zig").STATE_MAX) return error.TooManyStates;
         try self.states.ensureTotalCapacityPrecise(allocator, n_non_init);
 
         // Parse each state
@@ -199,6 +203,14 @@ pub const Model = struct {
                 try self.states_by_name.put(slot.name, slot);
             }
         }
+
+        // Make the stable-pointer invariant machine-checkable: any future code
+        // that accidentally appends to self.states would reallocate the slice
+        // and silently invalidate every *State held by states_by_name and by
+        // Transition.to_state_ptr. The capacity reservation above sized the
+        // slice to exactly n_non_init; if items.len ever drifts from capacity,
+        // the invariant has been broken.
+        std.debug.assert(self.states.items.len == self.states.capacity);
 
         try self.finalize(allocator);
     }
--- a/packages/zig-core/src/ham/model.zig
+++ b/packages/zig-core/src/ham/model.zig
@@ -173,9 +173,14 @@ pub const Model = struct {
                 // never reallocates and the &slot pointer is stable.
                 const slot = self.states.addOneAssumeCapacity();
                 slot.* = State.init();
-                // Roll back the addOne if parse or put fails — pop() removes the
-                // slot so deinit doesn't try to free a partially-built State that
-                // is also handled by the local cleanup here.
+                // Roll back the addOne if parse or states_by_name.put fails.
+                // Order matters: `slot.deinit` must precede `pop`, because after
+                // pop the slot index is past states.items.len and a later
+                // model.deinit would not see it. Conversely, without pop, the
+                // slot would still live in states.items and model.deinit would
+                // re-deinit it (double free). If states_by_name.put is the
+                // failing call, no key was stored (put is failure-atomic), so
+                // freeing slot.name in deinit doesn't dangle a map entry.
                 errdefer {
                     slot.deinit(allocator);
                     _ = self.states.pop();
--- a/packages/zig-core/src/ham/model.zig
+++ b/packages/zig-core/src/ham/model.zig
@@ -32,9 +32,19 @@ pub const Model = struct {
     /// Track (alphabet). Owned by this Model.
     track: ?*Track,
 
-    /// All non-init states in model order. Each pointer is heap-allocated and owned.
-    states: std.ArrayListUnmanaged(*State),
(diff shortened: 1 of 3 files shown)


matsen (commented):
## Self-review pass (clean-code-reviewer + zig-code-reviewer)

Two review agents ran on this branch. Findings and dispositions:

### Blocking (fixed in 558c8c1a3)

- **STATE_MAX off-by-one regression** in `model.zig` parse(). Old code checked `if (items.len >= STATE_MAX)` immediately before each append, so the (N+1)-th append (items.len already == STATE_MAX) was the one that errored — i.e. exactly 500 non-init states allowed. My initial bulk-reservation rewrite checked `if (n_non_init >= STATE_MAX)` on the total count, capping at 499. Changed to `>` to restore the inclusive STATE_MAX bound. Academic for production HMMs (~30-50 states each) but a real behavior change.

### Suggested (addressed in 558c8c1a3)

- **Capacity-invariant assertion missing.** The stable-pointer invariant (`&states.items[i]` valid because we never grow past initial capacity) was documented in comments but had no machine-checkable guard — a future stray `try self.states.append(...)` would silently reallocate and invalidate every `*State` in `states_by_name` and every `Transition.to_state_ptr`. Added `std.debug.assert(items.len == capacity)` after the parse loop.
- **Duplicate iteration over `data.states.items`** — collected `state_names` and counted `n_non_init` in two separate passes; merged into one.
- **`emissionLogprobSeqs` "bit-identical" wording** was technically over-broad (IEEE NaN payloads can differ across paths). Rewritten to "value-identical for every input emissionLogprob can produce" with a note that `emissionLogprob` never returns NaN in practice.
- **`errdefer` ordering** at `model.zig:184` (`slot.deinit` before `pop`) was correct but only documented by effect, not by ordering requirement. Comment in 7f1565ab2 made the ordering invariant explicit.

### Acknowledged but not addressed

- **Bundling 13+19 with 9** — clean-code-reviewer suggested splitting on PR-hygiene grounds (changes have no structural relationship to the pointer flattening). Decision to bundle was made up front to avoid a second 30k validation gate (~1.5h per run); items 13+19 are parity-trivial and add minimal bisect surface. Not splitting now.
- **Parse-loop complexity tradeoff** — clean-code-reviewer pushed back on the "simpler" framing for `parse()` since the loop now has two structurally different branches. Updated PR body to acknowledge this is a tradeoff (deinit simplification real, parse complexity real) rather than a pure win.

### Confirmed safe

- Memory safety / use-after-free risks in flat-slice approach: no callers grow `Model.states` after parse; `HMMHolder` heap-allocates `*Model` and never moves it; only `parse()` ever appends; capacity reserved before any append.
- Iteration site coverage: all `for (self.states.items) |st|` callsites updated to `|*st|`; compiler caught the obvious cases and there's no place where a value-copy bug could compile silently.
- Errdefer correctness: `slot.deinit` then `pop()` correctly avoids both leak (slot stays in items.len, deinit catches it) and double-free (pop removes from items.len so deinit can't see it).

### Validation re-run

- Rung 0 (`zig build test`): pass
- Rung 1 (`partis-test.py --quick --zig`): pass

The fixes are comment + assertion + one operator change; no behavior change for any production model. Re-running the 30k gate would be overkill for this delta.

PR #356, opened 2026-05-01
fix hfrac bin-packing to preserve TCM component boundaries

s-canchi:
Bin-packing in `_merge_r1_subgroups_by_components()` treated round-1 sub-groups as the indivisible unit. This allowed TCM components to be split across bins, fragmenting families that TCM identified as related.

TCM components are now the indivisible unit for bin-packing:
- Components at or above `min_group_size` are emitted as separate bins
- Only tiny components below that threshold are bundled together
- Components exceeding the cap are emitted whole (never split)

Also adds summary logging showing the three-level reduction (round-1 sub-groups -> TCM components -> output bins).

`partis-test.py --quick` passes. Validated on 1M soto-bcr IGH: no family splits, max bin 41K, wall time 6:50.


psathyrella (approved):
looks good!

PR #361, opened 2026-05-04
zig: per-query and per-kset arena allocators (#342 item 4)

matsen:
## Summary

- Sub-PR for #342 item 4. Design comment: https://github.com/psathyrella/partis/issues/342#issuecomment-4370363878.
- Adds a per-query `ArenaAllocator` to `DPHandler` (backs `scratch_cachefo`, `paths`, `scores`, `per_gene_support`, `run()`-local maps) and a per-kset arena inside `runKSet` (backs region sub-seq clones, kset-local maps, and the chunk-cache temp trellis). `parent_allocator` is reserved for `Result` / `RecoEvent` and anything the caller reads after `run()` exits.
- `Trellis.traceback` takes an explicit `path_alloc` parameter so the chunk-cache-hit path can run the temp trellis on the per-kset arena without the path's items capturing that shorter-lived arena (would otherwise UAF when the kset arena resets).
- The previous `clear()` cleanup loops are kept as safety belts (no-op on the arena); the arena reset at the bottom of `clear()` is what actually reclaims memory.
- `run()` calls `if (clear_cache) self.clear();` as its first action so the arena reset doesn't invalidate scratch built for the current call (seqs container, only_genes).

## Test plan

- [x] **rung 0** — `zig build test` (Debug + ReleaseSafe) passes.
- [x] **rung 1** — `partis-test.py --quick --zig` passes under ReleaseFast and ReleaseSafe (latter is the use-after-free check called out in the design comment for the `parent`/`scratch` boundary).
- [x] **rung 2a** — `partis-test.py --no-simu --zig` matches topic-tip (`824129fea`); only pre-existing test-framework noise (stale ref times, the `3 only in ref version` naive_seqs delta which also appears on topic-tip).
- [x] **rung 2b** — `partis-test.py --paired --no-simu --zig` matches topic-tip; same pre-existing notes.
- [x] **rung 4** — 5k paired partition byte-identical vs `/tmp/zig-5k-baseline`:

  ```
  65f2720cbcc59cb2d3f89ad1f5a0b2cb  /tmp/zig-5k-arena/igh+igk/partition-igh.yaml
  7ad6b1f3b0b9833fb7d11240bdc332d6  /tmp/zig-5k-arena/igh+igk/partition-igk.yaml
  ```
  matches the baseline md5s exactly. Wall 4:55 vs 5:10 baseline (n=1, suggestive only); peak RSS 467 MB unchanged.

- [ ] **30k pre-merge gate** — pending. Plan: n=2 both sides on pax (this branch + topic tip `824129fea`), report wall + user-CPU + peak RSS. Per the design comment, n≥2 is required to call any sub-2% wall delta given #360's lesson.
- [ ] 50k validation — deferred per matsen, not part of this PR.

### DIFF
--- a/packages/zig-core/src/ham/dp_handler.zig
+++ b/packages/zig-core/src/ham/dp_handler.zig
@@ -56,12 +56,29 @@ pub const DPHandler = struct {
     args: *Args,
     gl: *GermLines,
     hmms: *HMMHolder,
-    allocator: std.mem.Allocator,
+
+    /// Caller-supplied allocator. Used **only** for allocations that
+    /// outlive the DPHandler — Result, RecoEvent and any owned strings
+    /// reachable from them (gene names on RecoEvent, naive_seq, deletion
+    /// keys, insertion seqs). See item 4 of issue #342 for the lifetime
+    /// table; any allocation that lands in `parent_allocator` rather than
+    /// `scratchAllocator()` is a deliberate cross-boundary write.
+    parent_allocator: std.mem.Allocator,
+
+    /// Per-query arena, lives for the full DPHandler lifetime. Backs every
+    /// internal scratch allocation: scratch_cachefo, paths, scores,
+    /// per_gene_support, and the run()-local maps (only_genes,
+    /// best_scores, total_scores, best_genes). On `clear()`,
+    /// reset(.retain_capacity) drops every dependent allocation in O(1)
+    /// and reuses the backing pages for the next call. Item 4, #342.
+    query_arena: std.heap.ArenaAllocator,
 
     /// scratch_cachefo_: gene → list of *TrellisEntry. Entries are
     /// individually heap-allocated for stable address (the trellis at each
     /// entry borrows `&entry.query_seqs` and that pointer must stay valid
-    /// across appends to the same gene's list).
+    /// across appends to the same gene's list). Address stability is
+    /// preserved under the per-query arena because arenas only grow —
+    /// past allocations are never relocated.
     scratch_cachefo: std.StringHashMapUnmanaged(std.ArrayListUnmanaged(*TrellisEntry)),
     /// paths_: gene → (KSet → TracebackPath)
     paths: std.StringHashMapUnmanaged(std.AutoHashMapUnmanaged(KSet, TracebackPath)),
@@ -82,7 +99,8 @@ pub const DPHandler = struct {
             .args = args,
             .gl = gl,
             .hmms = hmms,
-            .allocator = allocator,
+            .parent_allocator = allocator,
+            .query_arena = std.heap.ArenaAllocator.init(allocator),
             .scratch_cachefo = .{},
             .paths = .{},
             .scores = .{},
@@ -90,14 +108,31 @@ pub const DPHandler = struct {
         };
     }
 
+    /// Per-query scratch allocator. **Always re-derive at the use site** —
+    /// caching `arena.allocator()` in a struct field would silently break
+    /// after a DPHandler move, since the returned `Allocator.ptr` points
+    /// at the arena's address at call time. (Item 4, #342.)
+    pub inline fn scratchAllocator(self: *DPHandler) std.mem.Allocator {
+        return self.query_arena.allocator();
+    }
+
     pub fn deinit(self: *DPHandler) void {
         self.clear();
+        self.query_arena.deinit();
     }
 
     /// Clear all cached trellis data.
     /// Corresponds to C++ `DPHandler::Clear`.
+    ///
+    /// Under the per-query arena (item 4, #342), every individual `free` /
+    /// `destroy` / `deinit` below is a no-op — the arena reset at the end
+    /// reclaims everything in one shot. The explicit cleanup loops are kept
+    /// as documentation and as a safety belt: they make the ownership
+    /// graph readable, and if a future change ever pulls one of these
+    /// allocations off the arena and back onto a real allocator, the
+    /// existing `free` will keep doing the right thing.
     pub fn clear(self: *DPHandler) void {
-        const allocator = self.allocator;
+        const allocator = self.scratchAllocator();
 
         // Free scratch_cachefo. Each *TrellisEntry is heap-allocated and the
         // entry's trellis borrows from the inline `query_seqs`. Deinit the
@@ -141,6 +176,11 @@ pub const DPHandler = struct {
         while (pgit.next()) |entry| allocator.free(entry.key_ptr.*);
         self.per_gene_support.deinit(allocator);
         self.per_gene_support = .{};
+
+        // Reset the per-query arena: this is what actually reclaims memory
+        // (the loops above are no-ops on the arena). Retains capacity so
+        // the next run() in handleFishyAnnotations reuses the backing pages.
+        _ = self.query_arena.reset(.retain_capacity);
     }
 
     /// Run the DP for a single Sequence.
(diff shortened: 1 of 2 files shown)


matsen:
## 30k pre-merge gate + 50k validation — done

Ran on pax (32-core, 123 GB RAM), interleaved order arena-topic-arena-topic to control for order/thermal confounds. All output md5-identical to `/tmp/zig-30k-baseline` and `/tmp/zig-50kcoll-baseline`.

### 30k paired partition (n=2 each side)

| | Wall | User-CPU | Peak RSS |
|---|---|---|---|
| topic tip `824129fea` | 1:12:03, 1:13:52 (mean 1:12:57) | 26841.45 s, 26974.31 s (mean 26907.88 s) | 1.84 GB |
| **this branch** | 1:13:03, 1:14:46 (mean 1:13:54) | 26484.47 s, 26685.65 s (**mean 26584.56 s**) | 1.84 GB |
| Δ | **+57 s (+1.3%)** | **−323 s (−1.2%)** | unchanged |

Spreads (max − min): wall 103 s arena / 109 s topic, user 201 s arena / 133 s topic. The wall delta (+57 s) is well within the within-side spreads, so wall must be called as **noise**. **User-CPU at −1.2% is larger than the topic-side spread of 133 s** (−323 s vs ±67 s) — that's a real signal in the cleaner channel that #360 already established for sub-2% effects. Per the design comment's lower-bound prediction (~12 s recovered), this overshoots by 25× — so something more than just glibc-malloc round-trip elimination is contributing, plausibly fewer key-dupe/hash-insert pairs from `per_gene_support_this_kset` collapsing into arena bumps.

The order of runs (arena-1 first, topic-2 last) systematically favors the topic side on wall if there's any "machine warms up over time" effect — which would make the +1.3% wall an underestimate of the real wall delta in arena's favor. User-CPU is less sensitive to system-level order effects, so it's the more trustworthy signal.

RSS unchanged at the MB level. The design predicted ~3 MB savings per query peak; aggregate peak RSS doesn't change.

### 50k collision (n=1 on this branch only — no topic-tip 50k run)

| | Wall | User-CPU | Peak RSS |
|---|---|---|---|
| pinned `/tmp/zig-50kcoll-baseline` (main + #351) | 4:16:58 | — | **22.8 GB** |
| this branch | **4:02:37** (n=1) | 75120.54 s | **13.37 GB** |
| Δ vs pinned baseline | −14:21 (−5.6%) | — | **−9.4 GB (−41%)** |

Output md5-identical to `/tmp/zig-50kcoll-baseline` for both loci.

The −41% RSS at 50k was **predicted** in the cold-start handoff: "if the proportional 30k RSS drop holds, peak RSS at 50k could fall from 22.8 GB toward ~14 GB" — and it did, landing at 13.4 GB. **This is mostly attributable to #355 (item 6, −39% at 30k), not to this PR's arenas** — the topic tip baseline at 30k is already 1.84 GB (down from 3.0 GB pre-#355). Without a topic-tip 50k run for direct comparison, this PR's marginal 50k RSS contribution can't be isolated; expected to be flat or sub-1% based on the 30k measurement. Headline matters anyway: 50k now fits comfortably in 16 GB hosts.

### Summary

- ✅ Byte-identical output at 30k (n=2) and 50k (n=1) — all four 30k runs and the one 50k run hit `/tmp/zig-30k-baseline` and `/tmp/zig-50kcoll-baseline` md5s exactly.
- User-CPU **−1.2% at 30k**; wall noisy (+1.3% within spread); RSS unchanged.
- 50k completes in 4:02 with 13.4 GB peak — comfortably within the predicted post-topic-cumulative range. No regression from arena.

Cumulative topic vs `main` (`d40bc889a`) baseline at 30k: wall 1:31:11 → 1:13:54 mean (**−18.9%, n=2**), RSS 3.0 GB → 1.84 GB (**−39%**). At 50k: 22.8 GB → 13.4 GB (**−41%**).

Ready for review.

matsen:
## Review pass — fixes pushed (`803f0e41d`); also a correction to the perf framing in [the previous comment](https://github.com/psathyrella/partis/pull/361#issuecomment-4374516037)

Ran the bip pre-merge review (zig-code-reviewer + surprising-conclusion-skeptic). Findings + actions below.

### Code review fixes (pushed in `803f0e41d`)

1. **Blocking — `handleFishyAnnotations` use-after-reset.** The inner `self.run(... clear_cache=true)` call resets `query_arena` as its first action, which would reclaim `naive_seq`'s `.name` / `.header` / `.undigitized` / `.seqq` backing bytes before the inner DP loop reads `seqq[i]` in `emissionLogprobSeqs`. Reviewer correctly noted the function is currently dormant (`pub` but unused — the C++ originals at `bcrham.cc:94` and `glomerator.cc:740` have the corresponding `HandleFishyAnnotations` calls commented out), but it's intended for future activation. Fixed by allocating `naive_seq` on `parent_allocator`.

2. **Advisory — `Trellis.allocator` move-invariant.** Trellises stored in `scratch_cachefo` capture `dph.scratchAllocator()` whose `Allocator.ptr` aliases `&dph.query_arena`; moving the DPHandler after a Trellis is built would dangle every stored `allocator.ptr`. All three call sites (`bcrham.zig:92`, `glomerator.zig:800,917`) bind `dph` to a stack `var` and never reassign, so the invariant holds — but it wasn't documented. Added a doc comment on the field.

3. **Advisory — `getSubSeqs` cleanup.** Dropped the unused `self` parameter; it's now a free function within the struct scope rather than a method with `_ = self;`.

Skipped advisories: orphaned-path-on-`gp.put`-overwrite (the type system doesn't enforce single-insertion per `(gene,kset)` but the runtime logic does, and reviewer agreed it's harmless arena waste at worst); `kset_arena` accounting nit (parent_allocator-backed is fine).

### Perf claim correction

The skeptic flagged two specific framings in [the previous comment](https://github.com/psathyrella/partis/pull/361#issuecomment-4374516037) as overstated. Both critiques are correct:

- **The "overshoots design lower bound by 25×" claim is apples-to-oranges.** The design's 12 s lower bound was a **wall** prediction (3×10⁸ allocs × 50 ns × 80%). The 323 s observed delta is **user-CPU**. Dividing by parallelism (CPU% ≈ 600%) gives ~54 s wall-equivalent — comfortably *inside* the predicted 12–480 s wall band, not 25× above it. **Withdraw** the "something more than glibc-malloc round-trip elimination is contributing" speculation that rested on this miscomparison; the result is consistent with the design's predicted mechanism without invoking extra effects.

- **The order/warming argument is post-hoc and not data-supported.** Within-side trends went *opposite* directions: arena-1 < arena-2 on user-CPU but topic-1 > topic-2. A simple monotonic warming/cooling effect would predict same-direction trends. The skeptic's parsimonious explanation — variable interactive load on pax across the day — fits the data better. **Withdraw** the claim that "topic ran later on average so the wall delta is an underestimate of arena's advantage."

Revised perf statement (replacing the bullet on "User-CPU at −1.2% is larger than the topic-side spread"):

> 30k user-CPU: arena 26585 s mean vs topic 26908 s mean (n=2 each). Sign reliable across both pairs (per-pair deltas −490 s and −156 s, both negative). Magnitude uncertain at n=2 — paired t with df=1 gives t≈−1.93 (95% critical ≈±12.7), so the underlying effect is plausibly anywhere in roughly **[−0.3%, −1.8%]** with the central estimate around −1.2%. **Cite as "≈1% user-CPU savings, sign reliable, magnitude uncertain at n=2,"** not as a settled −1.2%. The wall-equivalent (~54 s after CPU% normalization) sits inside the design's 12–480 s prediction band — consistent with the predicted glibc-round-trip mechanism, no surprise residual.

The other claims hold up: byte-identical output (independently re-md5'd by the skeptic), wall +1.3% within noise call (defensible), 50k attribution (honest about prior PRs), peak RSS unchanged. Skeptic's overall verdict: **"Check first" — broad direction credible, headline magnitude soft, no merge blockers from the perf side.**

### What's not changing

- Validation ladder is still green: rung 0 + rung 1 re-run after fixes (`partis-test.py --quick --zig`). The merged paths (`run` / `runKSet` / `fillTrellis` / `fillRecoEvent`) are unchanged in behavior; the only behavioral fix is in dead code (`handleFishyAnnotations`). 30k byte-equality and 50k byte-equality from the previous comment still apply to commit `f03be4b49`.
- For follow-up cumulative-impact accounting (e.g., topic-cumulative-vs-`main` summary in the issue), use **"~1% user-CPU"** with a wide CI rather than a settled −1.2%. If the precise magnitude matters for any downstream decision, n=3 each side is ~3 hours and would tighten the CI substantially.

PR #362, opened 2026-05-04
fix simulation UID hash collision at scale

s-canchi:
#### Simulation UID collision at 137K events

Running paired simulation with 137K events (`--n-sim-events 137000 --paired-loci`) hits a UID collision during paired merging. The per-locus simulation files write successfully, but `check_for_cross_locus_duplicates` crashes on a duplicate UID within a single locus.

Observed on two different seeds:
- seed 1: `7f20fc4f2c` duplicated in igh
- seed 11: `3b25054b58` duplicated in igl

Changing the random seed did not fix it.

The collision comes from `uidhashstr` in `event.py` truncating to 10 hex characters. At 137K events this hash space is too small to avoid collisions reliably. The subcluster annotation hash collision fix (`4a5f6fe83`) already increased `max_len` to 20 for that call site.

This PR applies the same fix to simulation UID generation in `event.py:99` and `event.py:107`.


psathyrella:
Looks good, thanks for catching!

PR #363, opened 2026-05-05
zig: small cleanups post-arena (#342 items 5, 15, 22)

matsen:
Bundle of three small post-arena cleanups against the long-lived topic branch `342-zig-perf-cleanup`. Items 5, 15, and 22 from issue #342, plus a note that item 10 was already done. Each item is mostly a code-quality / redundancy cleanup with a small expected wall-time bonus — wall is expected to be at noise level since item 4 (#361) already collapsed the underlying allocator round trips into arena bumps.

This PR is framed as redundancy / code-quality, not a perf headline. Per recurring lesson on this issue (PRs #360, #361), wall is too noisy below ~2% to call without n≥2 30k runs each side; this change's predicted effect is well under that threshold.

### Item 5 — `DPHandler.initCache` triple gene-name dupe → single shared key
Three `dupe(u8, gene)` calls collapsed to one shared key across `scratch_cachefo`, `paths`, and `scores`. The three maps share one lifetime on `query_arena`, so shared ownership is safe; `scores` is named the canonical owner so the existing `clear()` free pattern continues to do the right thing if these maps ever move off the arena. `ensureUnusedCapacity` + `putAssumeCapacity` keeps the puts infallible after the dupe (no errdefer hazard).

The pre-existing comment "Three separate key copies are needed because each HashMap owns its key independently" predated the per-query arena and is now removed.

### Item 15 — `LexicalTable.replaceLogProbs` alloc/free cycle → @memcpy on stable buffer
`setLogProbs` now allocates `original_log_probs` to the same size as `log_probs` and reuses it as a scratch buffer. `replaceLogProbs` and `unReplaceLogProbs` become `@memcpy` operations instead of free+dupe pairs, so the per-query rescale/un-rescale cycle (~50k calls per 5k-seq run × 4 f64 = 1.6 MB of per-query allocator traffic) no longer hits the allocator at all.

The `allocator` parameter on both functions is retained for API symmetry with `setLogProbs` (silenced via `_ = allocator;`) — dropping it would cascade through `Emission` and `State` signatures for no real benefit.

### Item 22 — `per_gene_support_this_kset` borrows gene keys from `only_genes`
The local map duped each gene name on insert, then freed on defer (~20k dupe/free pairs per query). The keys it received were already arena-duped slices into `only_genes`, which lives on the per-query arena (`scratchAllocator()`) and strictly outlives this map (per-kset arena). Borrow the slice directly; both the dupe and the defer-free loop go away.

### Item 10 — already done
The issue body claimed `StateBitset` was `[STATE_MAX]bool` (1 byte/bool, 500 B). Actual code at `state.zig:27` has been `pub const StateBitset = std.StaticBitSet(STATE_MAX);` since `6599319a1b` (Erick, 2026-03-16), which predates this issue. Marking item 10 done in the issue tracker (no diff in this PR).

## Validation

Standard iteration gate (per the cold-start handoff at issue-comment-4360930507) — all parity-safe by construction, since none of these touch arithmetic, traversal order, or any state read by the trellis:

- **rung 0** `zig build test` — pass
- **rung 1** `partis-test.py --quick --zig` — pass
- **rung 2a** `partis-test.py --no-simu --zig` — pass
- **rung 2b** `partis-test.py --paired --no-simu --zig` — pass after `uv pip install -e .` (initial run hit the stale `--max-ccf-fail-frac 0.15` threshold; reinstall bumped the venv to dev64 with the upstream 0.25 threshold)
- **rung 4** 5k-seq paired partition byte-equality vs `/tmp/zig-5k-baseline`:
  ```
  igh: 3797 events identical
  igk: 3802 events identical
  OK: identical
  ```

5k wall was 4:54 / 466 MB peak RSS vs the pinned 5:10 / 467 MB baseline (n=1 each, not a controlled comparison — within wall noise).

**30k pre-merge gate skipped for this PR.** Standard policy is n=2 each side at 30k for sub-PRs, but the design's expected wall delta is well below the n=2 30k resolution (~1–2% per #360 / #361 lessons), and rung-4 byte-equality already covers the parity gate. Running 30k would burn ~5 hours of compute to confirm "no detectable wall change" — not informative for this PR's scope. If a reviewer prefers the 30k anyway, happy to run it.

Refs #342.

### DIFF
--- a/packages/zig-core/src/ham/bcr_utils/hmm_holder.zig
+++ b/packages/zig-core/src/ham/bcr_utils/hmm_holder.zig
@@ -106,7 +106,7 @@ pub const HMMHolder = struct {
             var git = gene_set.iterator();
             while (git.next()) |entry| {
                 const m = try self.get(entry.key_ptr.*);
-                m.unRescaleOverallMuteFreq(self.allocator);
+                m.unRescaleOverallMuteFreq();
             }
         }
     }
--- a/packages/zig-core/src/ham/dp_handler.zig
+++ b/packages/zig-core/src/ham/dp_handler.zig
@@ -819,9 +819,10 @@ pub const DPHandler = struct {
         var regional_total: std.AutoHashMapUnmanaged(Region, f64) = .{};
         defer regional_total.deinit(allocator);
         // per_gene_support_this_kset borrows its keys from `only_genes`'s
-        // arena-duped strings (item 22 of #342). `only_genes` lives on the
-        // per-query arena (`scratchAllocator()`) for the entire `run()`
-        // call, strictly outliving this map (per-kset arena).
+        // arena-duped strings (item 22 of #342). `only_genes`'s key bytes
+        // live on the per-query arena (`scratchAllocator()`); they outlive
+        // any single `runKSet` call (per-kset arena dies first), and in
+        // fact persist until the next `clear()` resets the arena.
         var per_gene_support_this_kset: std.StringHashMapUnmanaged(f64) = .{};
         defer per_gene_support_this_kset.deinit(allocator);
 
--- a/packages/zig-core/src/ham/dp_handler.zig
+++ b/packages/zig-core/src/ham/dp_handler.zig
@@ -79,10 +79,17 @@ pub const DPHandler = struct {
     /// across appends to the same gene's list). Address stability is
     /// preserved under the per-query arena because arenas only grow —
     /// past allocations are never relocated.
+    ///
+    /// The string keys are NOT owned by this map — they are shared with
+    /// `paths` and `scores`, which are populated by the same `initCache`
+    /// call (item 5 of #342). `scores` owns the duped key bytes; `clear()`
+    /// frees keys only when iterating `scores` to avoid a triple-free.
     scratch_cachefo: std.StringHashMapUnmanaged(std.ArrayListUnmanaged(*TrellisEntry)),
-    /// paths_: gene → (KSet → TracebackPath)
+    /// paths_: gene → (KSet → TracebackPath). Keys are borrowed (see
+    /// `scratch_cachefo` doc-block); the canonical owner is `scores`.
     paths: std.StringHashMapUnmanaged(std.AutoHashMapUnmanaged(KSet, TracebackPath)),
-    /// scores_: gene → (KSet → f64)
+    /// scores_: gene → (KSet → f64). Owns the gene-name keys shared with
+    /// `scratch_cachefo` and `paths` (item 5 of #342).
     scores: std.StringHashMapUnmanaged(std.AutoHashMapUnmanaged(KSet, f64)),
     /// per_gene_support_: gene → best full-annotation log-prob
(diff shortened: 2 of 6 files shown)


matsen:
## Review pass — fixes pushed in `2e1e2bb4f`

Ran the bip pre-merge review (zig-code-reviewer + automated checks). Findings + actions below.

### Code-review fixes (pushed in `2e1e2bb4f`)

1. **`setLogProbs` non-atomic two-allocation sequence.** Original code freed `log_probs` and `original_log_probs` up front, then allocated. A second-alloc OOM would leave the struct with a stale `original_log_probs` slice header pointing at freed memory (caught later in `deinit` as a use-after-free / double-free). Fixed by allocating both new buffers first, then swapping them in only after both succeeded.

2. **`replaceLogProbs` state corruption on `BadNormalization`.** Original code did the two `@memcpy` calls before checking the normalization sum, so a `BadNormalization` return left `log_probs` overwritten with the bad distribution. Moved the normalization check ahead of both copies — the function is now atomic w.r.t. its error return. Added a regression test asserting `logProb(0)` is unchanged after a `BadNormalization`.

3. **`_ = allocator;` anti-pattern on `replaceLogProbs` / `unReplaceLogProbs`.** Reviewer made a strong case that keeping the unused parameter "for API symmetry" misleads readers about the function's contract and propagates dead arguments through the rescale chain. Dropped `allocator` from `LexicalTable` and `Emission` `replaceLogProbs` / `unReplaceLogProbs`, plus `State.unRescaleOverallMuteFreq` and `Model.unRescaleOverallMuteFreq` (which were only forwarding). `State.rescaleOverallMuteFreq` keeps its allocator parameter — still used for `self.emission.logProbs(allocator)`.

4. **Imprecise lifetime comment on `per_gene_support_this_kset` borrow.** Reviewer pointed out that `only_genes`'s keys actually live until the next `clear()` resets the per-query arena, not "for the entire `run()` call" as written. The borrow is still safe (longer lifetime, not shorter); reworded the comment.

### Skipped advisories

- Reviewer initially flagged a potential `clear()` resilience issue if `initCache` got partial state on `ensureUnusedCapacity` failure. Withdrew on second pass — the new structure is genuinely atomic (capacity reservation has no observable side effect on count; failures before `dupe` leave nothing to clean up; failures after `dupe` cannot occur since the three `putAssumeCapacity` calls are infallible).

- Pre-existing `zig fmt --check` violation at line 8 of both files (blank line after doc comment) — out of scope per the cold-start handoff ("Don't add new violations; don't bundle a fmt sweep into perf PRs"). My changes did not introduce new fmt violations.

### Validation re-run after fixes

All rungs re-run on `2e1e2bb4f`:

- rung 0: `zig build test` ✓
- rung 1: `partis-test.py --quick --zig` ✓
- rung 2a: `partis-test.py --no-simu --zig` ✓
- rung 2b: `partis-test.py --paired --no-simu --zig` ✓
- rung 4: 5k paired partition byte-identical to `/tmp/zig-5k-baseline` (3797 igh + 3802 igk events). Wall 4:44 / RSS 466 MB (n=1, not a controlled comparison).

PR #367, opened 2026-05-05
fix simulation UID collision with per-event counter

s-canchi:
Supersedes #362 (hash length increase alone was insufficient).

The root cause of simulation UID collisions at 137K+ events is that `irandom` values are drawn from int32, making collisions near-certain at 52K+ events. When two events get the same `irandom` and produce identical rearrangement parameters and sequences (e.g., two singletons using the same common V gene), their UIDs are identical regardless of hash length.

Fix: add a class-level event counter to the hash input in `event.py`, guaranteeing unique UID inputs per event within a process.

Validated: three paired simulations at 137K events each completed without UID collisions (previously failed on every attempt).

### DIFF
--- a/bin/partis
+++ b/bin/partis
@@ -275,7 +275,8 @@ def run_simulation(args):
 
     n_per_proc_list, work_fnames = [], []
     if args.n_procs == 1:
-        make_events(args.n_sim_events, args.workdir, args.outfname, [random.randint(0, numpy.iinfo(numpy.int32).max) for _ in range(args.n_sim_events)], heavy_chain_events=heavy_chain_events)
+        retry_gap = 20000  # must exceed max retry count in recombinator.combine() (currently 10000) so per-event retry seeds don't overlap with the next event's range
+        make_events(args.n_sim_events, args.workdir, args.outfname, [args.random_seed * retry_gap * args.n_sim_events + ievt * retry_gap for ievt in range(args.n_sim_events)], heavy_chain_events=heavy_chain_events)
     else:
         cmdfos = run_sub_cmds(working_gldir)
 
--- a/bin/partis
+++ b/bin/partis
@@ -185,7 +185,7 @@ def run_simulation(args):
             clist[0] = utils.find_cmd(clist[0])
             utils.replace_in_arglist(clist, '--n-procs', '1')
             clist.append('--im-a-subproc')
-            utils.replace_in_arglist(clist, '--random-seed', str(args.random_seed + iproc))
+            utils.replace_in_arglist(clist, '--random-seed', str(args.random_seed + iproc * args.n_sim_events))  # shift by n_sim_events so per-proc seed ranges (each proc uses [seed, seed+n_sub_events)) don't overlap
             utils.replace_in_arglist(clist, '--workdir', get_workdir(iproc))
             utils.replace_in_arglist(clist, '--outfname', get_outfname(iproc))
             utils.replace_in_arglist(clist, '--n-sim-events', str(n_sub_events))
@@ -275,8 +275,8 @@ def run_simulation(args):
 
     n_per_proc_list, work_fnames = [], []
     if args.n_procs == 1:
-        retry_gap = 20000  # must exceed max retry count in recombinator.combine() (currently 10000) so per-event retry seeds don't overlap with the next event's range
-        make_events(args.n_sim_events, args.workdir, args.outfname, [args.random_seed * retry_gap * args.n_sim_events + ievt * retry_gap for ievt in range(args.n_sim_events)], heavy_chain_events=heavy_chain_events)
+        # NOTE per-event sequential seeds (each event uses one seed slot since combine() seeds once and lets retries advance the RNG state). Multi-proc cross-proc uniqueness is handled by the iproc * n_sim_events shift above.
+        make_events(args.n_sim_events, args.workdir, args.outfname, [args.random_seed + ievt for ievt in range(args.n_sim_events)], heavy_chain_events=heavy_chain_events)
     else:
         cmdfos = run_sub_cmds(working_gldir)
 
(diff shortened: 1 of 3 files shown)


psathyrella:
Alternative worth considering: the underlying problem is at the seed-generation site, not the hash-input site. The `random.randint(0, int32max)` call in `bin/partis` (around the `make_events` invocation) draws from a space that's birthday-prone above ~65k events — but there's no actual reason those values need to be random draws. They only need to be (a) distinct across events within a process, (b) distinct across processes, and (c) deterministic given `--random-seed`. The "should be a random number, not just a seed" comment in `recombinator.py` is misleading: for the Mersenne Twister, seeds 0, 1, 2, … produce wildly different sequences, so the per-event RNG output is just as good with structured seeds.

A deterministic-counter scheme like

```python
random_ints = [args.random_seed * BIG + iproc * args.n_sim_events * RETRY_GAP + ievt * RETRY_GAP
               for ievt in range(args.n_sim_events)]
```

(with `RETRY_GAP` larger than `combine()`'s max retry count, so each event's retry seeds `initial_irandom + itry` can't overlap with the next event's range) would be **collision-free by construction** — within and across processes — without needing the `_event_counter` class state in `event.py`. It also removes a confusing piece of code (why is the seed itself random?) and ports cleanly to multi-proc.

Why this might be better than the per-event counter:
- Solves cross-process collisions too (per-process counter alone doesn't — proc A event 5 and proc B event 5 share `_event_id=5`; they're saved today only by independent irandom draws happening to differ, which is itself birthday-prone).
- One-line change in `bin/partis`, no new state in `event.py`.
- Doesn't change the hash-input scheme, so it's a smaller blast radius for downstream users with stored UIDs.

Trade-off: changes the irandom values, so ref UIDs change — but PR 367 already changes them, so the regeneration cost is the same.

Bumping the int range alone (to `uint32max` = `2^32-1`, the actual ceiling for `numpy.random.seed`) only doubles the birthday bound to ~92k, so it doesn't fix 137k. Going to 64-bit seeds would require switching from the legacy `numpy.random` global API to `numpy.random.default_rng(...)` — a much bigger refactor.

edit: hopefully we can also go back to length-10 simulation ids? Or at least as short as possible, since it really clutters up the debug output having them longer.

s-canchi:
Oh that's clever! I added your suggestion and reverted the hash length back to 10.

psathyrella:
Pushed `345acaef7` with a different fix. The deterministic-seed scheme I suggested earlier turns out to be broken — apologies for the misdirection.

**Why the previous suggestion was wrong.** The expression `args.random_seed * retry_gap * args.n_sim_events + ievt * retry_gap` overflows `numpy.random.seed`'s 32-bit cap. I confirmed by direct test: numpy raises `ValueError: Seed must be between 0 and 2**32 - 1` for any seed >= 2^32. With the default `--random-seed = int(time.time()) ~= 1.78×10^9`, this overflows immediately at any `n_sim_events`. Even with `--random-seed=2`, it crashes at 137k events. The scheme only happened to work in validation because the test seed was small.

**Why none of the in-place patches work.** Once you accept that retries need their own seed slots (because `try_to_combine` re-seeds on every entry), you need at least `n_procs × n_events × retry_gap` distinct seeds in 32-bit space. At 4 procs × 137k × 10000 = 5.5×10^9, that's larger than 2^32 — guaranteed collisions by pigeonhole, no arithmetic scheme can fix it. Tightening `retry_gap` is unsafe because `n_max_tries=10000` exists for a reason. Parent-coordinated rejection sampling is heavy plumbing. Restricting `--random-seed` breaks compat. Migrating to 64-bit via `numpy.random.default_rng` works but is a real refactor.

**What we did instead.** The retry-seed budget exists *only because* `try_to_combine` re-seeds on every call. Moving the seed up into `combine()` (seed once per event, let retries advance the RNG state naturally) removes the budget entirely — each event consumes exactly one seed slot regardless of retry count. This is the same pattern the inner `get_scratchline` retry loop already uses (no re-seed; the RNG state advances through `try_scratch_erode_insert` calls). The bit-budget requirement collapses from `n_procs × n_events × retry_gap` to just `n_procs × n_events`, four orders of magnitude smaller, and 32-bit space is comfortably sufficient.

Per-event seeds are now `args.random_seed + iproc * args.n_sim_events + ievt`. Reproducible, collision-free, no overflow.

**Verification.** Ran `partis simulate --simulate-from-scratch --n-sim-events 100` at `--n-procs {1, 10}` against three code states: dev (original `random.randint(0, int32max)`), the previous PR-367 commit, and this new commit. Parsed seeds from `combine (seed N)` debug output and UIDs/reco_ids from output yaml. Results:

- 1-proc: 100 events, seeds exactly `[42, 43, ..., 141]`, gap=1.
- 10-proc: 100 events, seeds in 10 disjoint per-proc ranges `[42..51], [142..151], ..., [942..951]`, gap=1 within proc / 91 between procs.
- 1-proc: 108 `try_to_combine` calls vs 100 distinct seeds → 8 retries reused the per-event seed (RNG state advanced).
- 10-proc: 106 vs 100 → 6 retries.
- All UIDs and reco_ids unique in both runs.

Test refs will need refreshing (UID strings change), but that was already true under your previous commit.

The previous `_event_counter` field in `event.py` from your first commit is no longer needed with this approach — happy to drop it in a follow-up if you agree.

s-canchi:
The fix in 345acaef7 is clean and well-verified. Seeding once per event in combine() and letting retries advance the RNG state naturally is the right approach. The _event_counter in event.py from the earlier commit can be dropped.

This also unblocks larger simulations (1M+ events) for hfrac threshold testing, since the seed budget is now just n_procs * n_sim_events rather than n_procs * n_events * retry_gap.

psathyrella (approved):
great!

PR #376, opened 2026-05-08
Fix Zig-vs-C++ 30k partition parity: stable_sort tied events (#375)

### LINKED ISSUE #375: Find and fix 30k Zig-vs-C++ cpp-mirror parity break
# Find and fix 30k Zig-vs-C++ cpp-mirror parity break

🤖

## Motivation

Diagnostic runs in PR #373 ([landed 2026-05-07](https://github.com/psathyrella/partis/pull/373)) revealed that the Zig and C++ backends produce different output on `/tmp/paired-paper-all-seqs.fa --n-max-queries 30000`. The break has been latent since the Zig integration (PR #339, merged 2026-04-29). Issue #342's validation ladder caps at rung 4 = 5k paired partition, and PR #339's Zig-vs-C++ checks ran on the canned `test/ref-results` fixture (~36 sequences) plus Duncan's `--n-final-clusters 1` seed-partition test on 32k igh-only sequences — neither catches a tiebreak-density-dependent divergence in unbounded paired partition at 30k.

This blocks confident topic→main acceptance for #366 because we can't distinguish "Zig perf change drifted output" from "Zig was already different at 30k." It also breaks the [cpp-mirror] contract that #366's framing depends on.

## Observed divergence

All numbers from `--paired-loci --paired-outdir … --infname /tmp/paired-paper-all-seqs.fa --n-max-queries 30000 --random-seed 1 --n-procs 10 --no-time-based-n-proc-reduction --dont-write-git-info`. Runs are deterministic across invocations of the same binary.

| state | sg-igh | sg-igk | p-igh | p-igk |
|---|---|---|---|---|
| Zig main `d40bc889a` (no Phase 0) | 9929 | 263 | 10802 | 10802 |
| C++ main `d40bc889a` (no Phase 0, no 3.2) | **9917** | **258** | **10750** | **10750** |
| Zig topic post-#373 (Phase 0 + 3.2) | 9922 | 258 | 10750 | 10750 |
| C++ topic post-#373 (3.2 only) | 9917 | 258 | 10750 | 10750 |

At main `d40bc889a`: 12-cluster delta in single-chain igh, 5 in single-chain igk, 52 per locus in paired partition. At topic post-revert: 5-cluster delta in single-chain igh; single-chain igk and paired partition cluster *counts* match but *memberships* differ (zig-compare reports `PARTITION DIFF (10750 vs 10750 clusters)`, indicating cluster-membership divergence even when counts agree).

Confounder note for any debugging session: Phase 0 (#368) introduced a small Zig-only output drift at 30k (Zig main 10802 → Zig Phase-0-only 10750) despite its `comptime`-gated counters claiming bit-equality. The Phase 0 drift narrows the Zig-vs-C++ gap (12 → 5 clusters in single-chain igh) but doesn't close it. Item 2.1 has zero output effect at 30k (verified empirically in #373). 3.2 has zero effect on either backend at 30k (verified empirically: C++ pre-3.2 ≡ C++ topic). The pre-existing parity break is upstream of all three #366 commits.

### Phase 1 — pin a minimal reproducer

Reduce the failing test to the smallest input that still exhibits Zig-vs-C++ divergence in single-chain igh cluster count. Start by halving `--n-max-queries` from 30000 until the divergence disappears, then bisect to pin the minimum N. Subset the input to a single droplet bracket if possible (`--seed-unique-id` won't reproduce because the divergence emerges from unbounded clustering).

Concrete starting commands (run on pax — both binaries are already built in topic post-merge):

```bash
source .venv/bin/activate
N=15000  # bisect down from here

# C++ baseline (uses the post-3.2 ham binary at packages/ham/bcrham)
PYTHONHASHSEED=0 python3 bin/partis partition --paired-loci \
  --paired-outdir /tmp/parity-repro-cpp \
  --infname /tmp/paired-paper-all-seqs.fa --n-max-queries $N \
  --random-seed 1 --n-procs 10 --no-time-based-n-proc-reduction \
  --dont-write-git-info

# Zig
PYTHONHASHSEED=0 python3 bin/partis partition --paired-loci \
  --paired-outdir /tmp/parity-repro-zig \
  --infname /tmp/paired-paper-all-seqs.fa --n-max-queries $N \
  --random-seed 1 --n-procs 10 --no-time-based-n-proc-reduction \
  --dont-write-git-info --zig

/tmp/zig-compare.py /tmp/parity-repro-cpp /tmp/parity-repro-zig
```

Note on the historical baseline: `/tmp/zig-30k-baseline-MAIN-d40bc889a-PREVIOUS/igh.fa` is 11,983,957 bytes vs the topic-state baselines at 11,910,574 bytes — generated against a slightly different input snapshot. Phase 1 should regenerate the C++ main baseline against the current `/tmp/paired-paper-all-seqs.fa` before bisecting; otherwise the historical baseline will appear "different" for non-DP reasons.

**Output:** a command line with the smallest N that produces a Zig-vs-C++ cluster-membership delta. Pin the minimum N and save the two output dirs as `/tmp/parity-repro-{cpp,zig}/` with a one-line note in this issue.

**Phase gate:** N < 30000, ideally N < 5000 or even N < 1000. A reproducer at N < 500 means rung-3 already exhibits the divergence and the issue is small-cluster-count rather than scale — investigate that hypothesis separately.

### Phase 2 — localize the first DP divergence

For the smallest reproducer cluster pair (the one that gets merged on one backend but not the other, or vice versa), dump per-position `forward_log_probs` from both backends and find the first position where they disagree. This requires running each backend with `--debug 1` or equivalent on the same input and comparing log lines.

Hot sites that are candidates for the divergence (in priority order based on prior parity-fix history in PR #339). For each, the Zig path is in `packages/zig-core/src/ham/`; the C++ path is in the `packages/ham` git submodule (currently pinned at `419ec56` on the ham `precision-fixes` branch — changes there land in `psathyrella/ham` first, then partis bumps the submodule pointer):

1. `addInLogSpace` — fast_softplus LUT vs glibc log/exp. Zig: `mathutils.zig:88`. C++: `packages/ham/include/mathutils.h` (header impl). LUT has bounded error 6.5e-9 per call; at 30k scale the cumulative error could flip merge tiebreaks.
2. Trellis `next_states` placement — already fixed once in PR #339 (commit `8f871e325` "fix trellis next_states placement to match C++ (inside i_st_prev loop)"). Zig: `trellis.zig` middleViterbiVals/middleForwardVals. C++: `packages/ham/src/trellis.cc`. Re-audit at scale.
3. HashMap iteration order in glomerator paths. Zig: `packages/zig-core/src/ham/glomerator/glomerator.zig`. C++: `packages/ham/src/glomerator.cc`. PR #339 sorted many but possibly missed some.
4. Cache I/O serialization (`stof` vs `stod` was a bug in PR #339, fixed in commit `9624b6334` "fix precision: stof→stod in C++ cache reads, use double arithmetic in GetMergedQuery; match in Zig"). Re-audit whether any double values still get f32-truncated through cache files at 30k where caches fill more.
5. `joint_mute_freq` arithmetic — Zig: `packages/zig-core/src/ham/glomerator/glomerator.zig:1273`. C++: `packages/ham/src/glomerator.cc` (`GetMergedQuery`). The f32 vs f64 fix at the C++ side (commit `0a03de5c5` reverted in `84b53d56a`) and Zig fix (`cec11fac8`). Re-audit whether the two backends still match at scale.

**Output:** the file path and line in the Zig + C++ DP code where the first divergence shows up. A specific operation (FP order, iteration order, type narrowing, etc.).

**Phase gate:** Phase 3 is blocked until Phase 2 produces a concrete root cause. If the per-position dumps don't show divergence (i.e., DP outputs match exactly but the partition still differs), the divergence is downstream in the glomerator — investigate hash-map iteration / cluster sort order.

### Phase 3 — fix on both backends as a [cpp-mirror] item

Apply the parity fix to both Zig and C++ in coordinated PRs per #366's [cpp-mirror procedure](https://github.com/psathyrella/partis/issues/366#issuecomment-4391141047):

1. Land C++ change in `packages/ham`, tag `cpp-tag`.
2. Bust ref-results (`bin/partis-test.py --bust-cache` + `--paired --bust-cache`).
3. Land Zig change against `cpp-tag`.
4. Re-pin `/tmp/zig-{500,5k,30k}-baseline` from new C++ output.

If the root cause is a Zig-only bug (e.g., a HashMap iteration order leak), only the Zig side needs to change. Either way, the gate after the fix is bit-equal output between Zig and C++ at 30k.

**Phase gate:** the new validation gate from Phase 4 passes.

### Phase 4 — add 30k Zig-vs-C++ gate to validation infrastructure

`bin/partis-test.py` is a 3-line shim; the body lives at `partis/scripts/partis_test.py` (1027 lines). The script organizes tests as a dict of named ptests (`cache-parameters-simu`, `partition-new-simu`, etc.), parameterized by `--quick`/`--paired`/`--slow`/`--zig`. There's no "rung" concept in the file; rungs 0–4 are the issue-#342 validation-ladder framing, not an infrastructure abstraction. The new gate should be a separate test that runs both backends, since `partis-test.py` is structured around running one backend at a time.

Implementation options (pick one in Phase 4):

(a) **Standalone script** at `bin/partis-30k-parity-test.sh` that runs both backends sequentially on `/tmp/paired-paper-all-seqs.fa --n-max-queries 30000`, then `zig-compare.py`s the two output dirs. Minimal — just a shell script alongside `bin/partis-test.py`.

(b) **New `--zig-parity` mode** in `partis_test.py` that runs each existing ptest twice (with and without `--zig`) and diffs outputs. More invasive but reuses the existing test harness.

(c) **CI job** that runs (a) on a schedule (nightly) rather than per-PR, since the gate takes ~2.5 hours wall.

Recommended: (a) for the immediate gate, optionally lift to (c) once stable.

The gate should be runnable in under 4 hours wall (the 30k partition takes ~50 min Zig / ~100 min C++ on pax = 32-core AMD Zen 5). Even if Phase 3 closes the current gap, the gate is needed to prevent regression.

**Phase gate:** the chosen implementation runs cleanly on topic post-fix and produces "OK: identical" across single-chain igh, single-chain igk, and paired igh+igk loci, comparing Zig to C++ output at the same commit.

## Files to modify

- **Phase 1**: shell scripts and notes only; no source changes.
- **Phase 2**: temporary debug instrumentation in Zig (`packages/zig-core/src/ham/{dp_handler,trellis,mathutils}.zig`, `packages/zig-core/src/ham/glomerator/glomerator.zig`) and C++ (`packages/ham/src/{dphandler,trellis,glomerator}.cc`). All instrumentation reverted before Phase 3.
- **Phase 3**: depends on the root cause. Most likely candidates: `packages/zig-core/src/ham/{trellis,dp_handler}.zig`, `packages/zig-core/src/ham/glomerator/glomerator.zig`, `packages/ham/src/{trellis,glomerator,dphandler}.cc` (the latter via a coordinated PR on `psathyrella/ham`, then a submodule pointer bump on the partis side).
- **Phase 4**: a new `bin/partis-30k-parity-test.sh` (or other implementation per the Phase-4 options above), wiring through `partis/scripts/partis_test.py` if option (b) is chosen.

### Phased acceptance (per phase)

- Phase 1: minimal reproducer command line gives different cluster count between Zig and C++ binaries; document on this issue.
- Phase 2: per-position log-prob dump shows first divergence at a specific (cluster, position) pair; document on this issue.
- Phase 3: post-fix, the Phase 1 reproducer gives bit-equal output between backends.
- Phase 4: gate passes on topic post-fix; deliberately introducing a small DP perturbation on one backend makes it fail (sanity check that the gate has teeth).

### Validation tests beyond the 30k byte-equality gate (added per #366 review)

The 30k gate alone leaves blind spots: it tests one input at one `--n-procs` setting at one random seed. Several cheap auxiliary tests would convincingly bound the parity space.

1. **Symmetry across `--n-procs`.** Run Zig at `--n-procs 1`, `--n-procs 5`, `--n-procs 10` on a 5k subset and assert the three outputs are bit-identical. This catches HashMap-iteration-order or subprocess-ordering bugs that aren't exposed at single `--n-procs`. Repeat for C++. Then assert Zig ≡ C++ at each `--n-procs`. Cost: 3 × ~3 min Zig + 3 × ~5 min C++ = ~25 min.

2. **Random-seed sweep at 5k.** Run 5k partition with `--random-seed 1..10` for both backends; all 10 must be Zig-vs-C++ bit-equal. Catches tiebreak-density-dependent divergences without paying the 30k cost. Cost: 10 × ~5 min Zig + 10 × ~7 min C++ ≈ 2 hours; can run in parallel.

3. **Known-answer 4-sequence cluster.** Hand-construct 4 sequences with a known correct partition under both backends (use the existing `test/ref-results/test/simu.yaml` 36-seq fixture as a starting point or a smaller bespoke fixture). Compute per-cluster forward log-prob with both backends; assert agreement to ≤1e-12 (well below 1e-9 LUT-error budget). Cost: <1 min.

4. **LUT-error injection (root-cause confirmation).** If Phase 2 implicates `addInLogSpace`, replace `fast_softplus` with glibc `log1p(exp(x))` behind a build flag (`-Dno-fast-softplus=true`) and rerun the 30k Zig-vs-C++ test. If the divergence vanishes, the root cause is confirmed as fast_softplus accumulation. Cost: ~50 min Zig rebuild + 30k run.

### Regression

- `bin/partis-test.py --paired --no-simu` continues to pass (currently passes on topic).
- `bin/partis-test.py --no-simu` continues to show 0/14 logprob diffs and 0/40 naive-seq diffs.
- The four landed perf items on topic (Phase 0 #368 + 3.2 #372 + 2.1 revert #373 + fast_softplus fix #374) continue to give bit-identical output post-fix.

## Diagnostic baselines (preserved)

On pax `/tmp/`, generated this session for reference:

- `/tmp/zig-30k-baseline-MAIN-d40bc889a-PREVIOUS` — Zig main, 30k
- `/tmp/cpp-30k-pre32` — C++ main-equivalent (post-revert C++ binary `/tmp/bcrham-glibc-old`), 30k
- `/tmp/zig-30k-phase0only` — Zig at e3a63a6bf (Phase 0 alone), 30k
- `/tmp/zig-30k-baseline-topic-glibc` — Zig topic-no-3.2, 30k
- `/tmp/zig-30k-baseline` — Zig topic-with-3.2 (pre-revert), 30k
- `/tmp/cpp-30k-test` — C++ topic-with-3.2, 30k
- `/tmp/zig-30k-revert-test` — Zig topic post-revert (#373), 30k
- `/tmp/zig-compare.py` — comparison harness

`zig-compare.py` reports both cluster-count and event-membership equality.

## Success criteria

1. Zig and C++ produce bit-identical paired-partition output at 30k on `/tmp/paired-paper-all-seqs.fa --n-max-queries 30000` — `zig-compare.py` reports "OK: identical" across single-chain igh, single-chain igk, and paired igh+igk.
2. The same Zig-vs-C++ bit-equality holds at 5k (`/tmp/zig-5k-baseline` ≡ C++ at the same commit) and at the canned `test/ref-results` fixture (regression check).
3. A rung-5 30k Zig-vs-C++ gate exists and runs in under 4 hours wall time on pax.
4. The root cause is documented (specific file + line + operation type) so future re-occurrences can be diagnosed faster.

## Scope boundaries

**In scope:**
- 30k unbounded paired partition Zig-vs-C++ bit-equality
- Adding 30k as a rung-5 gate
- Documenting the root cause

**Out of scope:**
- 50k-collision parity (separate effort; topic→main acceptance gate per #366)
- Phase 0 (#368) gating bug — the comptime gating doesn't fully eliminate counter side effects. The drift is small (Zig single-chain igh shifts 9929 → 9922, single-chain igk 263 → 258, paired igh+igk 10802 → 10750 per locus on the same input) but is orthogonal to the cpp-mirror question. The drift accidentally narrows the Zig-vs-C++ gap rather than widening it. Track separately as a Zig-only follow-up; don't conflate with this cpp-mirror issue.
- Performance work in #366 — this issue blocks confident #366 progress but is a separate effort.

## Depends on / Blocked by

- Depends on PR #373 (`revert-item-2.1`) and PR #374 (`fix-fast-softplus-oob`) — both landed 2026-05-07.
- Blocks: confident topic→main acceptance for #366 (per PERF-NEXT.md re-pin procedure step 4 — "Confirm Zig binary-equals C++ at the bridge commit").

## References

- PR #373 — `revert-item-2.1` (diagnostic that surfaced the parity break)
- PR #374 — `fix-fast-softplus-oob` (related bug found during PR #373 review)
- Issue #366 — `partis-zig-core: next-round performance levers`
- Issue #342 — `Zig backend: reduce memory usage and improve efficiency` (original Zig perf cleanup; validation ladder caps at rung 4 = 5k)
- PR #339 — `Zig drop-in replacement for bcrham and ig-sw` (Zig integration; parity validation history)
- Issue #366 comment 2026-05-07 — full diagnostic write-up: https://github.com/psathyrella/partis/issues/366#issuecomment-4399144858

matsen:
## Summary

Fixes the latent 30k Zig-vs-C++ paired-partition divergence (issue #375). Root cause: **`Result::Finalize`** in C++ uses `std::sort` (unstable introsort) over events whose `score_` field is stored as `float`. Two ksets whose f64 viterbi scores differ by ~1e-6 collapse to **identical f32 values**; the unstable sort then picks an algorithm-dependent tied event, which **disagrees with Zig's `std.sort.block` (stable) pdqsort**. Different best_event → different gene calls → different naive_seq → different hfrac → different cluster merges. The cascade only surfaces at >20k where enough float-tied ksets exist for the divergence to bite.

### C++ (`packages/ham`, submodule bump)

`packages/ham/src/bcrutils.cc` — replace `sort(events_) + reverse(events_)` with `stable_sort(events_, descending_by_score)`. Both backends now pick the **first-pushed tied event** (push order is the kset reverse-iteration `k_v=vmax-1..vmin, k_d=dmax-1..dmin`). Submodule bumped to `375-stable-sort-events @ d354b76`. Companion ham PR: psathyrella/ham#20.

### Zig (`packages/zig-core`)

- `src/ham/bcr_utils/result.zig` — switch `std.mem.sort` (unstable pdqsort) to `std.sort.block` (stable). Defensive: pdqsort happens to give the same answer as block sort on the workloads tested, but the contract should be stable.
- `src/ham/glomerator/glomerator.zig` (lines 470-491) — fix stale comments and switch cache-read `parseFloat(f32, ...)` → `parseFloat(f64, ...)` for `logprob` and `naive_hfrac` columns. C++ has used `stod` since ham `f41e055`; the Zig comment claimed "match C++ stof" was wrong. Has zero observable effect on output for the 30k workload (verified — Zig with f32 truncation produces byte-identical 30k output to Zig with f64), but is the correct round-trip behavior.
- `src/ham/text.zig` — same `parseFloat(f32, ...)` → `parseFloat(f64, ...)` for `floatify` (port of `Floatify`; defined but unused).

### Phase 4 — gate

- `bin/partis-30k-parity-test.sh` — runs both backends sequentially on 30k unbounded paired partition, then `bin/zig-compare.py`s the two output dirs (cluster-membership + per-event annotation field equality across single-chain igh, single-chain igk, paired igh+igk). Default N = 30000; configurable. ~2.5 h wall on pax. Exits non-zero on diff.
- `bin/zig-compare.py` — comparison harness (was at `/tmp/zig-compare.py` during issue #375 development, now checked in).

## Bisection (Phase 1)

| N | Zig vs C++ (before fix) | Notes |
|---|---|---|
| 5000 | identical | 3797 igh + 3802 igk events |
| 10000 | identical | 6176 igh + 6189 igk events |
| 15000 | identical | 7794 igh + 7830 igk events |
| 20000 | identical | 9017 igh + 9055 igk events |
| 25000 | **1 igh field diff** | query `4126151233240036126-igh`: cpp `d_3p_del=4 dj_insertion='A'` vs zig `d_3p_del=3 dj_insertion=''` |
| 30000 | 919 paired-igh + 919 paired-igk member diffs (cluster counts identical at 10750) | |

The 25k 1-event diff was the smallest reproducer and the lever for the root-cause analysis. With the same FASTA + identical parameter dir, both backends compute Viterbi log-prob `-282.396001` exactly and select the same best kset `(k_v=293, k_d=11)` and same gene calls `V=IGHV3-66*01, D=IGHD1-26*01, J=IGHJ5*02`, but pick different traceback paths through the D-J junction.

## Verification

- [x] **1-query repro** (post-fix): cpp and zig both produce `d_3p_del=3 dj_insertion=''` for the divergent query
- [x] **5k Zig vs C++** (post-fix): identical (regression check)
- [x] **30k Zig vs C++** (post-fix): identical (10750 igh + 10815 igk events match)
- [x] `bin/partis-test.py --paired --no-simu`: passes
- [x] `bin/partis-test.py --paired --no-simu --zig`: passes

## How the agent found this

The empirical-bisection ladder (5k/10k/15k/20k all bit-equal, 25k differs by exactly one query) made it possible to isolate a **single divergent query** and diff the per-region viterbi DP between backends with debug instrumentation. The DP tables themselves were bit-identical at every position — divergence emerged only at the `Result::Finalize` step where the per-kset events are sorted to pick the best. The proximate cause (f32-collapsed ties + unstable sort) is hidden by float storage and only matters at the kset level when two paths score within 1e-6.

## Scope notes

- **`[cpp-mirror]` per #366 procedure**: ham change must land in psathyrella/ham main first, then partis submodule bump. This PR's submodule pointer is at the local commit `d354b76` for review; replace with the upstream tag once landed.
- **`partis-test.py` ref-results may need busting**: the canned tests pass against the current ref-results (small workloads, no f32-tied ksets in the fixtures), but `--bust-cache` would be a defensive sweep — non-blocking for this PR.
- **Phase 0 (#368) drift remains orthogonal**: the issue's "out of scope" Zig-only Phase-0 comptime-counter side-effect is not addressed here. Track separately.

Closes #375

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/bin/partis-30k-parity-test.sh
+++ b/bin/partis-30k-parity-test.sh
@@ -29,10 +29,7 @@ N="${3:-30000}"
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 COMPARE="${REPO_ROOT}/bin/zig-compare.py"
 if [[ ! -x "${COMPARE}" ]]; then
-  COMPARE="/tmp/zig-compare.py"  # legacy location used during issue #375 development
-fi
-if [[ ! -x "${COMPARE}" ]]; then
-  echo "ERROR: cannot find zig-compare.py (looked in bin/ and /tmp/)" >&2
+  echo "ERROR: cannot find zig-compare.py at ${COMPARE}" >&2
   exit 2
 fi
 
@@ -45,20 +42,21 @@ CPP_DIR="${OUTBASE}-cpp"
 ZIG_DIR="${OUTBASE}-zig"
 
 run_backend() {
-  local backend="$1" outdir="$2" extra_flag="$3"
+  local backend="$1" outdir="$2"
+  shift 2
   rm -rf "${outdir}"
   echo ">> running ${backend} backend → ${outdir}"
   PYTHONHASHSEED=0 python3 "${REPO_ROOT}/bin/partis" partition --paired-loci \
     --paired-outdir "${outdir}" \
     --infname "${INFILE}" --n-max-queries "${N}" \
     --random-seed 1 --n-procs 10 --no-time-based-n-proc-reduction \
-    --dont-write-git-info ${extra_flag}
+    --dont-write-git-info "$@"
 }
 
 # C++ first (slower), then Zig — sequential keeps the host responsive at the
 # usual 10 procs. Either order works; the partition is deterministic.
-run_backend "C++"  "${CPP_DIR}" ""
-run_backend "Zig"  "${ZIG_DIR}" "--zig"
+run_backend "C++"  "${CPP_DIR}"
+run_backend "Zig"  "${ZIG_DIR}"  --zig
 
 echo ">> comparing"
 "${COMPARE}" "${CPP_DIR}" "${ZIG_DIR}"
--- a/bin/partis-30k-parity-test.sh
+++ b/bin/partis-30k-parity-test.sh
@@ -1,0 +1,64 @@
+#!/usr/bin/env bash
+# 30k Zig-vs-C++ partition parity gate (issue #375).
+#
+# Runs an unbounded paired-loci partition with the Zig and C++ backends on the
+# same 30k input, then compares cluster membership and per-event annotation
+# fields. Exits 0 if identical, 1 otherwise. The 5k partis-test.py rung doesn't
+# exercise enough cache round-trips to surface tiebreak-density-dependent
+# divergences (#375 was latent until 30k).
+#
+# Usage:
+#   bin/partis-30k-parity-test.sh [INFILE [OUTBASE [N]]]
+#
+# Defaults:
+#   INFILE = /tmp/paired-paper-all-seqs.fa
+#   OUTBASE = /tmp/parity-30k-test
+#   N = 30000
+#
+# Wall budget on pax (32-core Zen 5):
+#   ~50 min Zig + ~100 min C++ + a few min compare = ~2.5 h total when sequential.
+#   Backends run sequentially to keep the host responsive; bump --n-procs in
+#   this script to parallelize within a backend.
+
+set -euo pipefail
+
+INFILE="${1:-/tmp/paired-paper-all-seqs.fa}"
+OUTBASE="${2:-/tmp/parity-30k-test}"
+N="${3:-30000}"
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+COMPARE="${REPO_ROOT}/bin/zig-compare.py"
+if [[ ! -x "${COMPARE}" ]]; then
+  COMPARE="/tmp/zig-compare.py"  # legacy location used during issue #375 development
+fi
+if [[ ! -x "${COMPARE}" ]]; then
+  echo "ERROR: cannot find zig-compare.py (looked in bin/ and /tmp/)" >&2
+  exit 2
+fi
+
+if [[ ! -f "${INFILE}" ]]; then
+  echo "ERROR: input file ${INFILE} does not exist" >&2
+  exit 2
+fi
+
+CPP_DIR="${OUTBASE}-cpp"
+ZIG_DIR="${OUTBASE}-zig"
+
+run_backend() {
+  local backend="$1" outdir="$2" extra_flag="$3"
+  rm -rf "${outdir}"
+  echo ">> running ${backend} backend → ${outdir}"
+  PYTHONHASHSEED=0 python3 "${REPO_ROOT}/bin/partis" partition --paired-loci \
+    --paired-outdir "${outdir}" \
+    --infname "${INFILE}" --n-max-queries "${N}" \
+    --random-seed 1 --n-procs 10 --no-time-based-n-proc-reduction \
+    --dont-write-git-info ${extra_flag}
+}
+
+# C++ first (slower), then Zig — sequential keeps the host responsive at the
+# usual 10 procs. Either order works; the partition is deterministic.
+run_backend "C++"  "${CPP_DIR}" ""
+run_backend "Zig"  "${ZIG_DIR}" "--zig"
+
+echo ">> comparing"
+"${COMPARE}" "${CPP_DIR}" "${ZIG_DIR}"
--- a/bin/zig-compare.py
+++ b/bin/zig-compare.py
@@ -3,13 +3,14 @@
 
 Usage: zig-compare.py <baseline_dir> <new_dir>
 
-Per issue #342 validation harness. Exit 0 if identical, 1 otherwise.
+Companion to bin/partis-30k-parity-test.sh (issue #375 cpp-mirror gate).
+Exit 0 if identical, 1 otherwise.
 """
 import json
 import os
 import sys
 
-sys.path.insert(0, '/home/matsen/re/partis')
+sys.path.insert(0, os.path.join(os.path.dirname(os.path.abspath(__file__)), '..'))
 from partis.clusterpath import ClusterPath
 
 
@@ -45,8 +46,10 @@ def diff_pair(base, new):
             print(f'{locus}: PARTITION DIFF ({len(bpart)} vs {len(npart)} clusters)')
             n_diffs += 1
             continue
-        bd = json.load(open(bp))
-        nd = json.load(open(np_))
+        with open(bp) as f:
+            bd = json.load(f)
+        with open(np_) as f:
+            nd = json.load(f)
         bk = {tuple(sorted(e['unique_ids'])): e for e in bd['events']}
         nk = {tuple(sorted(e['unique_ids'])): e for e in nd['events']}
         if set(bk) != set(nk):
--- a/bin/zig-compare.py
+++ b/bin/zig-compare.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+"""Diff two partis paired-loci output dirs for byte-equality on partition + annotations.
+
+Usage: zig-compare.py <baseline_dir> <new_dir>
+
+Per issue #342 validation harness. Exit 0 if identical, 1 otherwise.
+"""
+import json
+import os
+import sys
+
+sys.path.insert(0, '/home/matsen/re/partis')
+from partis.clusterpath import ClusterPath
+
+
+def best_partition(path):
+    cpath = ClusterPath()
+    cpath.readfile(path)
+    return sorted(tuple(sorted(c)) for c in cpath.partitions[cpath.i_best])
+
+
+FIELDS = ('naive_seq', 'v_gene', 'd_gene', 'j_gene',
+          'v_3p_del', 'd_5p_del', 'd_3p_del', 'j_5p_del',
+          'vd_insertion', 'dj_insertion', 'cdr3_length')
+
+
+def diff_pair(base, new):
+    n_diffs = 0
+    for locus in ('igh', 'igk', 'igl'):
+        bp = f'{base}/igh+igk/partition-{locus}.yaml'
+        np_ = f'{new}/igh+igk/partition-{locus}.yaml'
+        if not os.path.exists(bp):
+            bp = f'{base}/igh+igl/partition-{locus}.yaml'
+            np_ = f'{new}/igh+igl/partition-{locus}.yaml'
+        if not os.path.exists(bp):
+            print(f'{locus}: no baseline file (skipping)')
+            continue
+        if not os.path.exists(np_):
+            print(f'{locus}: MISSING in new ({np_})')
+            n_diffs += 1
+            continue
+        bpart = best_partition(bp)
+        npart = best_partition(np_)
+        if bpart != npart:
+            print(f'{locus}: PARTITION DIFF ({len(bpart)} vs {len(npart)} clusters)')
+            n_diffs += 1
+            continue
+        bd = json.load(open(bp))
+        nd = json.load(open(np_))
+        bk = {tuple(sorted(e['unique_ids'])): e for e in bd['events']}
+        nk = {tuple(sorted(e['unique_ids'])): e for e in nd['events']}
+        if set(bk) != set(nk):
+            only_b = set(bk) - set(nk)
+            only_n = set(nk) - set(bk)
+            print(f'{locus}: EVENT-KEY DIFF (only in base: {len(only_b)}, only in new: {len(only_n)})')
+            n_diffs += 1
+            continue
+        field_diffs = 0
+        for k in bk:
+            for f in FIELDS:
+                if bk[k].get(f) != nk[k].get(f):
+                    field_diffs += 1
+                    if field_diffs <= 3:
+                        print(f'{locus} {k[:1]}...: {f} diff: {bk[k].get(f)} vs {nk[k].get(f)}')
+        if field_diffs:
+            print(f'{locus}: {field_diffs} FIELD DIFFS across {len(bk)} events')
+            n_diffs += 1
+        else:
+            print(f'{locus}: {len(bk)} events identical')
+    return n_diffs
+
+
+def main():
+    if len(sys.argv) != 3:
+        sys.exit(__doc__)
+    base, new = sys.argv[1], sys.argv[2]
+    n = diff_pair(base, new)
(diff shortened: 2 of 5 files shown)


matsen:
Squash-merged into the 366-zig-perf-next topic branch as 081d65cd3, since this PR was the blocker for that topic per #375. Companion ham fix landed on `375-stable-sort-events` (rebased onto `419ec56` fast_softplus, now at `35bf4a9`); ham PR #20 left open for separate landing to ham main.

PR #397, opened 2026-07-27
size simulation UID hash length by event count to avoid collisions at scale

s-canchi:
Simulation UIDs use a 10-char (40-bit) hash. The deterministic per-event seeds from #367 removed same-input collisions, but the 10-char hash still birthday-collides at roughly 1M or more sequences (about 1-2 duplicate UIDs per dataset), which crashes the paired combine (`check_for_cross_locus_duplicates`).

This adds `--uid-hash-len` (auto by default):

- Unset resolves from the total `--n-sim-events`: 20-char at or above 100000 events, else 10-char.
- The resolved value is forwarded through the per-locus and per-proc subprocesses into `event.py` UID generation, and through the paired combine (`merge_heavy_light_trees` joint uid and reco_id).
- Small simulations keep readable 10-char UIDs; large simulations get collision-free 20-char UIDs.
- Builds on #367 and does not change the seed fix.
- `merge_heavy_light_trees` is simulation-only (single caller), so inference is unaffected.

#### Validation

- Auto-resolution: 100000 events gives 20-char, small gives 10-char.
- Single-locus and paired: default 10-char; `--uid-hash-len 20` gives 20-char UIDs and reco_ids through the combine.
- `partis-test.py --quick` passes with no regression.

### DIFF
--- a/bin/partis
+++ b/bin/partis
@@ -42,6 +42,7 @@ import partis.disjointgrouper as disjointgrouper
 from partis.parametercounter import ParameterCounter
 from partis.corrcounter import CorrCounter
 from partis.waterer import Waterer
+import partis.event as event
 
 # ----------------------------------------------------------------------------------------
 def strip_paired_args(clist, args):
@@ -137,6 +138,8 @@ def assemble_groups_action(args):
 
 # ----------------------------------------------------------------------------------------
 def run_simulation(args):
+    if args.uid_hash_len is None:  # single-locus: resolve from n-sim-events (paired resolves globally in run_all_loci)
+        args.uid_hash_len = event.UID_HASH_LEN_LONG if args.n_sim_events >= event.UID_HASH_LEN_THRESHOLD else event.UID_HASH_LEN_SHORT
     # ----------------------------------------------------------------------------------------
     def run_sub_cmds(working_gldir):
         # ----------------------------------------------------------------------------------------
@@ -189,6 +192,7 @@ def run_simulation(args):
             utils.replace_in_arglist(clist, '--workdir', get_workdir(iproc))
             utils.replace_in_arglist(clist, '--outfname', get_outfname(iproc))
             utils.replace_in_arglist(clist, '--n-sim-events', str(n_sub_events))
+            utils.replace_in_arglist(clist, '--uid-hash-len', str(args.uid_hash_len))  # forward resolved (global) hash len
             if working_gldir is not None:
                 utils.replace_in_arglist(clist, '--initial-germline-dir', working_gldir)
             if args.allele_prevalence_fname is not None:
@@ -815,6 +819,7 @@ def run_all_loci(args, ig_or_tr='ig'):
                 clist.append('--crash-on-duplicate-uids')  # single-chain subprocesses in paired mode should crash on duplicates
                 if args.action == 'simulate':
                     clist += ['--choose-trees-in-order', '--outfname', getofn(ltmp, lpair=lpair)]
+                    utils.replace_in_arglist(clist, '--uid-hash-len', str(args.uid_hash_len))  # forward resolved (global) hash len
                     if args.input_simulation_treefname is None:
                         clist += ['--input-simulation-treefname', getofn(ltmp, trees=True, lpair=lpair)]
                     if args.paired_correlation_values is not None and not utils.has_d_gene(ltmp):
@@ -1062,7 +1067,7 @@ def run_all_loci(args, ig_or_tr='ig'):
             print('%s: synchronizing heavy and light chain simulation trees and rewriting output files in %s/' % (utils.color('blue', '+'.join(lpair)), getodir(lpair=lpair)))
             antns_to_remove = []
             for hline, lline in paircluster.get_antn_pairs(lpair, lpfos):
-                treeutils.merge_heavy_light_trees(hline, lline, use_identical_uids=False)
+                treeutils.merge_heavy_light_trees(hline, lline, use_identical_uids=False, uid_hash_len=args.uid_hash_len)
                 if args.remove_nonfunctional_seqs:
                     rm_nonfunc_seqs(antns_to_remove, hline, lline)
             if len(antns_to_remove) > 0:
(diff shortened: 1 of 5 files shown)


psathyrella:
It seems maybe better to not have the extra default value of 10, as long as you already have the global constants in event.py? CC review following.

psathyrella (approved):
Reviewed via the group's PR-review checklist (manual control-flow read + clean-code and code-reuse agents). **Correct and safe to merge as-is** — the notes below are quality/robustness, not blockers.

### Correctness — sound
- Resolution runs once from the **global** `--n-sim-events` at the top of each simulate path (`bin/partis:141` single-locus, `:1511` paired), *before* the value is forwarded into subprocess arg lists (`:195`, `:822`), so subprocesses get a resolved int and don't re-resolve from their smaller per-locus counts. `replace_in_arglist` appends the flag when absent, so forwarding is robust.
- No downstream code assumes a 10-char UID (no `[:10]` / `len==10`) — UIDs are opaque keys, so 10→20 is safe. Small sims keep 10-char UIDs, so `--quick` / reference results don't regress.

### Worth deciding before merge: threshold is on events, but collisions are on sequences
Birthday collisions depend on total **sequences** (events × leaves/event), but the switch gates on **event** count.
- 40-bit hash: expected first collision ≈ **1.3M sequences**; P(collision) ≈ 0.37 at 1M, ≈1.0 at 5M.
- 100k events maps to 300k seqs (3 leaves) up to 5M seqs (50 leaves).
- So e.g. **99k events × 50 leaves ≈ 5M seqs stays at 10-char and is near-certain to hit the exact crash this fixes.** With default small families (geometric, mean ~5) it's an edge case, but the guard is defeatable.

Options: (a) accept it and note in the `--uid-hash-len` help that the auto-threshold assumes modest family sizes; or (b) threshold on estimated total sequences (`n_sim_events × expected leaves`). Leaning (a) since the leaf estimate isn't cleanly available at resolution time — but the help text should say so.

### One optional refactor that dissolves several small findings
The resolution expression is duplicated byte-for-byte at `bin/partis:141` and `:1511`, and the value `10` now lives unlinked in three places: `utils.uidhashstr`'s `max_len=10` default, `event.UID_HASH_LEN_SHORT=10`, and `merge_heavy_light_trees`'s hardcoded `uid_hash_len=10` default (bare literal because `treeutils` can't import `event` — the import only goes `event → treeutils`).

Putting the three constants in **`utils.py`** (next to `uidhashstr`) instead of `event.py`, plus a one-line resolver:
```python
def resolve_uid_hash_len(n_sim_events):
    return UID_HASH_LEN_LONG if n_sim_events >= UID_HASH_LEN_THRESHOLD else UID_HASH_LEN_SHORT
```
lets both `bin/partis` sites call `utils.resolve_uid_hash_len(...)`, makes `merge_heavy_light_trees`'s default `utils.UID_HASH_LEN_SHORT`, and drops the new `import partis.event as event` added to `bin/partis` just to reach three integers. `utils` is already imported everywhere with no circular-import issue — same "resolve once, reuse" pattern as the existing `get_n_events`.

### Minor
- `event.py:16` `uid_hash_len or UID_HASH_LEN_SHORT` uses truthiness where the rest of the diff uses explicit `is None` (harmless — 0 is never valid — but inconsistent).
- `recombinator.py:194` `getattr(self.args, 'uid_hash_len', None)` guards a case that can't happen (single caller always sets it), though it matches an existing idiom — leave or simplify to `self.args.uid_hash_len`.
- `bin/partis:822` forwards without `insert_after=` unlike the sibling `--n-sim-events` forward in the same block; fine if flag order doesn't matter, a one-word comment saying so would match the neighbor.

No scientific claim (skeptic skipped) and no new tests (the collision only bites at 1M+ seqs, so a regression test is impractical — worth noting the scale path is manually verified only).

s-canchi:
Thank you for the review. All the code items are applied; I kept the event-based threshold and
documented the caveat instead.

#### Applied

- Constants and `resolve_uid_hash_len()` moved to `utils.py` next to `uidhashstr`; both `bin/partis`
  sites call it, dropping the duplicated expression and the `import partis.event as event`.
- One exception: left `uidhashstr`'s own `max_len=10` default alone, with a NOTE. Its other callers are
  sw-cache filenames (`partitiondriver.py:124`), cluster hashes and the dup-uid suffix, so coupling them
  means a change to the simulation length silently renames sw-cache files.
- `merge_heavy_light_trees`: dropped the default entirely rather than naming it, since there is one
  caller and it always passes the value. Easy to change if you meant just the constant.
- Minors: truthiness to plain assignment; `getattr` to `self.args.uid_hash_len` (that case cannot happen,
  `Recombinator` is only built in `make_events` inside `run_simulation`, below the resolution);
  `insert_after='--choose-trees-in-order'` to match the sibling forward.

#### Threshold

Your arithmetic is right and the guard is defeatable, but the hash protects two things that scale
differently:

- reco ids: one per event, so an event threshold is exact.
- uids: events x leaves, so it assumes modest family sizes.

An `n_sim_events * n_leaves` estimate only helps the explicit-family-size case. The data-mimicking sims
use `--n-leaf-distribution hist`, where there is no `--n-leaves` to read, so it degrades to current
behavior exactly where it matters most while looking more principled than it is. At about 1.5 leaves/event
there, the event threshold errs safe. So: help-text note rather than a change.

#### Correction, plus the case that is actually silent

- I had assumed a single-chain collision silently desyncs truth. Wrong: `seqfileopener` renames via
  `choose_non_dup_id`, prints two warnings, and keys `reco_info` on the renamed uid, so
  `get_partition_from_reco_info` builds truth from the same names the input has.
- `reco_id` is the silent one. Same truncation, and `get_partition_from_reco_info` groups by it, so two
  colliding events merge into one true cluster with nothing checking
  (`check_for_cross_locus_duplicates` only looks at uids). P about 0.19 per donor at 685K events and 10
  chars. Negligible for cluster-count-vs-truth, but another reason to switch, and the part the event
  threshold gets exactly right.

`partis-test.py --quick` clean, and single-locus plus paired verified at both hash lengths through the
combine.

PR #394, opened 2026-07-24
single-chain refinement pipeline: ha-repartition and partition-refine

s-canchi:
Under `--disjoint-groups`, large CDR3 groups are partitioned with the fast vsearch heuristic, which over-merges similar-but-distinct families (light-chain single-chain purity is often ~0.05 to 0.10). This PR adds a post-partition refinement pipeline that recovers most of that accuracy at vsearch speed, to produce high-purity single-chain clusters for downstream parent-child and lineage inference.

- Off by default; requires `--disjoint-groups`; output is single-chain only (the paired combine is skipped, see below).
- Two stages in a fixed order: `--ha-repartition`, then `--partition-refine`. The order is enforced in code and cannot be inverted; only the full stack is characterized here.

#### Pipeline

```
  CDR3-length groups  (--disjoint-groups)
          |
   vsearch base partition          fast, but over-merges families
          |
   --ha-repartition                per-cluster HMM keep-or-split: undo over-merges
          |
   --partition-refine   step 1     validated split   (fix over-splits from a shared naive)
                        step 2     naive merge        (recover fragmented families)
                        step 3     shared-descent split (fix residual over-merges)
          |
   single-chain partition          (paired clustering skipped by guard)
```

Refine always runs after ha-repartition and consumes its output (it falls back to the raw vsearch partition only if ha-repartition was not run). ha-repartition-alone and refine-directly-on-vsearch are possible but uncharacterized.

#### Stages

- `--ha-repartition`: for each vsearch cluster of at least three sequences, rerun the full HMM clustering as an isolated keep-or-split decision and keep it only where it splits an over-merged cluster. Independent per cluster, so fast.
- `--partition-refine` step 1 (validated split): a Jaccard test on shared mutations proposes a split, and the per-sequence inferred naive validates it (fragments with agreeing naives are re-merged). A fragment with no SW naive to validate against is kept merged rather than split on unvalidated evidence.
- step 2 (naive merge): clusters with near-identical naives are merged only if their somatic-mutation fingerprints also agree, recovering families that were fragmented.
- step 3 (shared-descent split): described next.

#### What changed in step 3 (and why)

[...]

PR #400, opened 2026-08-13
Add a convergence-aware veto to the heavy naive-identity split

s-canchi:
#### Add a convergence-aware veto to the heavy naive-identity split

The heavy split proposes on exact naive identity and its only protection against breaking up a real family is shared mutations between members. When a family has few or no mutations there is nothing for that check to fire on, so the split can fragment it.

This adds a second protection that does not depend on mutations: if several members share the same inferred rearrangement and that rearrangement is rare enough that generating it repeatedly is implausible, the cluster is not split. The rarity threshold is the median generation probability of the rearrangement distribution the parameter directory defines, estimated once per locus.

#### Effect, 200k igh, against simulation truth

| cell | correct pairs | wrong pairs | family retention |
|---|---|---|---|
| d1 | **+1,158** | +6 | 0.7018 to 0.7034 |
| d4 | **+114** | 0 | 0.8666 to 0.8675 |
| d3 | **+44** | 0 | 0.9121 to 0.9129 |

Worst-cluster distinct-family count, the number of clusters holding 11 or more families, and the retention of families with 100 or more members are unchanged on every cell.

On the 50k convergent-recombination samples, where every true family is a singleton and any intervention would be damage, the veto fires on **0 of 9,149** multi-member clusters. Convergent sequences reach similar naives through different rearrangements, so they do not share the exact rearrangement the veto keys on.

The veto is on by default. Passing 0 for the minimum shared count disables it.

### DIFF
--- a/bin/partis
+++ b/bin/partis
@@ -1579,7 +1579,7 @@ def run_all_loci(args, ig_or_tr='ig'):
         print('%s %s with disjoint groups:' % (utils.color('blue_bkg', 'partition'), utils.color('blue', ltmp)))
         if args.partition_refine and utils.has_d_gene(ltmp):  # check up front for an incomplete parameter dir
             import partis.partition_refinement as partition_refinement
-            partition_refinement.validate_pgen_tables(getpdir(ltmp))
+            partition_refinement.validate_rearr_prob_tables(getpdir(ltmp))
 
         # step 1: CDR3 grouping
         manifest_path = '%s/%s' % (disjoint_dir, disjointgrouper.MANIFEST_FNAME)
--- a/bin/partis
+++ b/bin/partis
@@ -1579,7 +1579,7 @@ def run_all_loci(args, ig_or_tr='ig'):
         print('%s %s with disjoint groups:' % (utils.color('blue_bkg', 'partition'), utils.color('blue', ltmp)))
         if args.partition_refine and utils.has_d_gene(ltmp):  # check up front for an incomplete parameter dir
             import partis.partition_refinement as partition_refinement
-            partition_refinement.validate_rearr_prob_tables(getpdir(ltmp))
+            partition_refinement.validate_length_tables(getpdir(ltmp))
 
         # step 1: CDR3 grouping
         manifest_path = '%s/%s' % (disjoint_dir, disjointgrouper.MANIFEST_FNAME)
--- a/bin/partis
+++ b/bin/partis
@@ -240,7 +240,8 @@ def run_partition_refine_jobs_action(args):
     naive_thr = partition_refinement.estimate_locuswide_threshold(specs) if len(specs) > 0 and utils.has_d_gene(args.locus) else None
     stop = len(specs) if args.job_count is None else args.job_start + args.job_count
     specs = specs[args.job_start : stop]
-    partition_refinement.run_jobs(specs, naive_threshold=naive_thr, overwrite=args.overwrite, locus=args.locus)
+    partition_refinement.run_jobs(specs, naive_threshold=naive_thr, overwrite=args.overwrite, locus=args.locus,
+                                  parameter_dir=args.parameter_dir)
     if len(specs) > 0:  # per-slice done-marker so internal run_cmds can detect completion
         open(partition_refinement.bundle_marker_fname(disjoint_dir, args.job_start), 'w').close()
     print('  refined %d groups' % len(specs))
--- a/bin/partis
+++ b/bin/partis
@@ -1577,6 +1577,9 @@ def run_all_loci(args, ig_or_tr='ig'):
             return
         disjoint_dir = '%s/disjoint-groups/%s' % (getodir(single_chain=True), ltmp)
         print('%s %s with disjoint groups:' % (utils.color('blue_bkg', 'partition'), utils.color('blue', ltmp)))
+        if args.partition_refine and utils.has_d_gene(ltmp):  # check up front for an incomplete parameter dir
+            import partis.partition_refinement as partition_refinement
+            partition_refinement.validate_pgen_tables(getpdir(ltmp))
 
         # step 1: CDR3 grouping
         manifest_path = '%s/%s' % (disjoint_dir, disjointgrouper.MANIFEST_FNAME)
--- a/partis/partition_refinement.py
+++ b/partis/partition_refinement.py
@@ -339,32 +339,33 @@ def cross_shared_counts(frag_of_uid, uid_muts):
 # ----------------------------------------------------------------------------
 
 # members that must share the modal rearrangement for the veto to apply
-PGEN_VETO_MIN_SHARED = 4
-PGEN_MIN_PROB = 1e-9  # probability floor for a value the tables never saw
-PGEN_SAMPLE_DRAWS = 1000000  # rearrangements drawn to estimate the veto's cutoff
-PGEN_SAMPLE_SEED = 1  # fixed, so one parameter dir always gives one cutoff
+REARR_PROB_VETO_MIN_SHARED = 4
+REARR_PROB_MIN = 1e-9  # probability floor for a value the tables never saw, well below the
+# smallest probability any observed table entry can reach
+REARR_PROB_SAMPLE_DRAWS = 1000000  # rearrangements drawn to estimate the veto's cutoff
+REARR_PROB_SAMPLE_SEED = 1  # fixed, so one parameter dir always gives one cutoff
 
 # (name, column) per factorisation term; file and conditioning columns come from utils
-_PGEN_TABLE_COLUMNS = [('v', 'v_gene'), ('d', 'd_gene'), ('j', 'j_gene'),
-                       ('v_3p_del', 'v_3p_del'), ('d_5p_del', 'd_5p_del'),
-                       ('d_3p_del', 'd_3p_del'), ('j_5p_del', 'j_5p_del'),
-                       ('len_vd', 'vd_insertion'), ('len_dj', 'dj_insertion')]
+_REARR_PROB_TABLE_COLUMNS = [('v', 'v_gene'), ('d', 'd_gene'), ('j', 'j_gene'),
+                             ('v_3p_del', 'v_3p_del'), ('d_5p_del', 'd_5p_del'),
+                             ('d_3p_del', 'd_3p_del'), ('j_5p_del', 'j_5p_del'),
+                             ('len_vd', 'vd_insertion'), ('len_dj', 'dj_insertion')]
 
 
-def _pgen_table_specs():
+def _rearr_prob_table_specs():
     """(name, file, [varying column, conditioning columns...]) per factorisation term."""
     from partis import utils
     specs = []
-    for name, column in _PGEN_TABLE_COLUMNS:
+    for name, column in _REARR_PROB_TABLE_COLUMNS:
         cols = [column] + utils.column_dependencies[column]
         specs.append((name, utils.get_parameter_fname(column_and_deps=cols), cols))
     return specs
 
 
-PGEN_TABLE_SPECS = _pgen_table_specs()
+REARR_PROB_TABLE_SPECS = _rearr_prob_table_specs()
 
 
-def read_pgen_tables(parameter_dir):
+def read_rearr_prob_tables(parameter_dir):
     """Read the P(r) factorisation's count tables from <parameter_dir>/hmm (the locus-level
     parameter dir) and normalise each within its conditioning variable. Returns
     {name: {(value, conditioned-on...): probability}}, or None if <parameter_dir> is None.
@@ -375,14 +376,14 @@ def read_pgen_tables(parameter_dir):
     if not os.path.isdir(tdir):  # also accept the hmm dir itself
         tdir = parameter_dir
     tables = {}
-    for name, fname, cols in PGEN_TABLE_SPECS:
+    for name, fname, cols in REARR_PROB_TABLE_SPECS:
         counts = defaultdict(dict)
         try:
             with open('%s/%s' % (tdir, fname)) as tfile:
                 for row in csv.DictReader(tfile):
                     counts[tuple(row[c] for c in cols[1:])][row[cols[0]]] = float(row['count'])
         except (IOError, OSError, KeyError, ValueError) as terr:
-            raise Exception('couldn\'t read pgen table %s/%s (%s)' % (tdir, fname, terr))
+            raise Exception('couldn\'t read rearr-prob table %s/%s (%s)' % (tdir, fname, terr))
         tables[name] = {}
         for cond, cfo in counts.items():
             tot = sum(cfo.values())
@@ -402,7 +403,7 @@ def get_rearrangement(feat):
     return tuple(feat['vdj']) + tuple(vals)
 
 
-def rearrangement_lpgen(rearr, tables):
+def rearrangement_lprob(rearr, tables):
     """log10 P(r): the probability of generating the single inferred rearrangement (from
     get_rearrangement()), as the product of the parameter dir's marginals times 0.25 per inserted
     base. Comparable only against values from the same factorisation."""
@@ -411,8 +412,8 @@ def rearrangement_lpgen(rearr, tables):
              ('v_3p_del', (str(v_3p_del), v)), ('d_5p_del', (str(d_5p_del), d)),
              ('d_3p_del', (str(d_3p_del), d)), ('j_5p_del', (str(j_5p_del), j)),
              ('len_vd', (str(len_vd), d)), ('len_dj', (str(len_dj), j))]
-    lpgen = sum(math.log10(max(tables[name].get(key, 0.), PGEN_MIN_PROB)) for name, key in terms)
-    return lpgen + (len_vd + len_dj) * math.log10(0.25)
+    lprob = sum(math.log10(max(tables[name].get(key, 0.), REARR_PROB_MIN)) for name, key in terms)
+    return lprob + (len_vd + len_dj) * math.log10(0.25)
 
 
 def get_cluster_rearrangement(uids, uid_rearr_features):
@@ -456,7 +457,7 @@ def _draw_conditioned(rng, choices, groups, genes, n_draws):
     conditioned on a gene is keyed by an integer, and a gene the table never saw draws zero at
     the probability floor."""
     vals = np.zeros(n_draws)
-    lprobs = np.full(n_draws, math.log10(PGEN_MIN_PROB))
+    lprobs = np.full(n_draws, math.log10(REARR_PROB_MIN))
     for igene, idxs in groups.items():
         cond = (genes[igene],)
         if cond not in choices:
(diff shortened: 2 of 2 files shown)


psathyrella:
CC tells me there aren't any tests for partition refinement? If true, that seems bad to me. I think I remember we decided not to turn it on in the default partis-test.py tests yet, but it seems quite dangerous to me to have all these agent-made code changes without a ton of agent-style code checks like in our other repos (e.g. [loris-experiments](https://github.com/matsengrp/loris-experiments/tree/main/tests)). Arguably partis ought to have had more of those, instead of or in addition to its more correctly-style tests. But I do just think agents are so much more error prone (perhaps more accurately, they neither understand nor care about big picture goals) that they require the more detailed style of tests (which, luckily, they are also good at implementing).

CC review below. I'm still working through the tex, but once I finish that I'll have a better understanding of the whole refinement workflow, and will be better able to comment on the correctness of this new piece. It sounds like a good idea! But also seems quite complicated, and some of CC's comments below seem worrisome.

psathyrella:
> Claude Code:

Thanks for this -- the A/B harness is the cleanest one you've sent. Both arms call the shipped `refine_partition` on the same input partition object with only `pgen_veto_min_shared` differing, the adaptive thresholds come out identical in both arms (d1 0.0223), and the plumbing reaches the veto on both the standalone and integrated paths. The clustering gains are real and fairly measured.

The problems are with the mechanism and the safety argument rather than the numbers. I reproduced each of the three below against your parameter dirs before writing this.

#### The rarity clause admits half of everything

Fraction of observed rearrangements at or below the sampled cutoff, measured on the exact dirs the runs used:

| dir | tau | all observed (count-weighted) | among those observed >= 4x |
|---|---|---|---|
| d1 | -16.26 | 0.501 | 0.467 (1846/3955) |
| d3 | -16.64 | 0.500 | 0.465 (513/1104) |

That is expected from the construction: tau is the median of the same distribution the draws come from, so `P[log10 P(r) <= tau] = 0.5` by definition, and the observed rearrangements track it. The clause discards about half the eligible vetoes on a criterion that is close to orthogonal to the question being asked.

The prose describes it as "rare enough that generating it repeatedly is implausible", which is a statement about a tail, not a median. Formalised as a null it would be `K ~ Poisson(N * P(r))`, veto iff `P[K >= n_shared] < alpha` -- the same shape as the light-chain shared-descent test already in this module. The implemented test depends on neither `N` nor `n_shared` beyond the fixed `>= 4` gate. Over 200k model draws the *most probable* rearrangement has log10 P = -6.75, so at N = 200,000 its expected independent count is 0.036 and `P[K >= 4] ~ 1e-7`: the stated principle is satisfied by the entire distribution, and the median is doing something else.

So the operative threshold is `PGEN_VETO_MIN_SHARED = 4`, which is the constant that is neither swept nor identified as the threshold anywhere in the PR.

One related correction: the `sample_pgen_median` docstring says the cutoff is "a generative property of the locus, so it needs no pass over the data and does not depend on how the input happened to be clustered". The tables are counted by partis annotating the same sample, so tau does move with the sample -- the d1/d3/d4 spread (-16.26 / -16.64 / -16.58) is that dependence.

#### The 0 of 9,149 result cannot fire

The 50k parameter dir holds **49,968 distinct rearrangements for 49,984 sequences**, collapsed onto the veto's own 9-tuple. Nothing repeats even twice, let alone four times. `n_shared >= 4` is unsatisfiable there, so the veto could not fire regardless of pgen, cluster composition or convergence. The cluster sizes agree: of 2,643 multi-member clusters only 60 reach size 4.

The stated reason -- convergent sequences reaching similar naives through different rearrangements -- is true, but it is a property of an all-singleton scratch simulation rather than a discrimination property of the veto. As it stands the safety test and the benefit claim sit in disjoint regimes, and the result carries no information about behaviour on data where the veto can fire.

Two related things I could not resolve from disk. `validate-013-30k50k.log` ends in `KeyError: 'logprob'` from `clusterpath.readlines`, so the run that would have measured actual clustering damage never finished. And `validate-013-50k-firing.py:28` does `json.load(open(a.base_partition))`, but the partitions under `refine-singlechain/` are YAML and no JSON partition exists in that tree, so the script as committed cannot run against those files; I could not find its output anywhere under `hmm-split-expt/`. If you ran it against a converted file, could you send the command and output?

#### The reported benefit comes from 55 clusters

The log reports 1,537 / 444 / 722 held. Differencing the counters against the veto-off arm shows where those came from, including the unprinted `ctr['no_proposal']` bucket, which accounts for the residual:

| cell | held | from single_naive | from no_proposal | from rejected | from **accepted** |
|---|---|---|---|---|---|
| d1 | 1,537 | 1,234 | 163 | 99 | **41** |
| d3 | 444 | 414 | 17 | 9 | **4** |
| d4 | 722 | 682 | 21 | 9 | **10** |

Only the `accepted` column is a cluster that would otherwise have been split; the rest were kept whole either way. So `ctr['pgen_veto']` overstates the intervention by roughly 37x on d1, and the headline rests on 41 / 4 / 10 clusters from a single seed with no replicates. +1,158 correct pairs against +6 wrong is consistent with that -- refusing to split one cluster of size n contributes up to n(n-1)/2 pairs at once -- but it does mean the result is a small-sample one and should be reported as such.

#### Naming, and the #394 rejection

`pgen` here is not the published quantity. OLGA (Sethna 2019) and Murugan 2012 define `P_gen(sigma)` as a marginal over every recombination scenario consistent with an observed sequence; the sum is the expensive part and is what makes the quantity annotation-independent. This computes `P(r)` for the single scenario partis inferred, under an independence-factorised model fit by counting hard annotation calls. Defensible for what the veto asks, but the name invites a comparison it is not making -- either rename it to a rearrangement rarity score or state the distinction in the docstring.

Separately, #394 tested a pgen criterion and concluded "only ordinal information survives, so Pgen is out". This PR doesn't mention that. The new form does escape those specific failures, but it does so by dropping `N` and `n_shared`, which is to say by no longer asking a question about the repertoire. Worth saying so explicitly.

#### Code

Existing partis code covers three pieces of this:

- `utils.column_dependencies` (`utils.py:692`) already encodes the conditioning structure in `PGEN_TABLE_SPECS`, term for term, and `utils.get_parameter_fname` (`utils.py:2343`) generates all nine filenames. Building the specs from those instead of literals keeps them from drifting -- and the drift would be silent, since `read_pgen_tables` returns `None` on any mismatch and the veto just turns off. (The independence assumption itself is fine, and matches the HMM's own factorisation; it's the hand copy I'd change.)
- `recombinator.read_vdj_version_freqs` (`recombinator.py:241`) and `choose_vdj_combo` (`:469`) already draw complete rearrangements from the joint `all-probs.csv`, which carries the correlations the factorisation drops. Worth a line on why the factorised form was chosen over the joint the simulator itself samples from -- particularly since the d1/d3/d4 sims were generated by that same code path, so the generator and the scorer are using different models of the same thing.
- `recombinator.read_insertion_content` (`recombinator.py:102`) reads the real per-base insertion content and falls back to uniform 0.25 only when the data is absent. The hardcoded `0.25` at lines 408 and 480 is unconditional, and the content files are measurably non-uniform.

Other items:

- **No CLI arg.** The PR says passing 0 disables the veto, but `pgen_veto_min_shared` is only a Python kwarg -- there's no argparse entry. A default-on change to splitting behaviour should have a command-line way off.
- **`_sampling_choices` divides by zero.** `read_pgen_tables` guards `tot > 0` and emits a 0.0 row, then `probs / probs.sum()` (line 432) produces NaN and `rng.choice` raises. `sample_pgen_median` is outside the try/except, so this crashes the run rather than disabling the veto as the docstring promises. Latent -- no real parameter file has zero-count rows, and `parametercounter` only writes what it observed -- but the guard exists one function up and not here.
- **The probability floor fails toward vetoing.** `tables[name].get(key, 0.)` floors a *missing* key at `PGEN_MIN_PROB`, so an unseen rearrangement (odd deletion length, an allele absent from the counts, a mis-annotation) scores -9 per missing factor and clears tau almost surely. Sampling never hits the floor, since it only draws table-present values, so this is evaluation-side only. Might be the direction you want; it isn't documented either way.
- **No majority condition.** `n_shared` is compared to an absolute 4 with no fraction-of-cluster requirement, unlike the rearrangement-majority guard already in this module, and uids with incomplete features are dropped from the count silently. On d1, 3,955 distinct rearrangements occur 4 or more times and carry 16.7% of the repertoire, so nothing bounds the size of the cluster a 4-member match can protect. What's the largest cluster vetoed on d1?
- **Tie-breaking.** `max(counts, key=counts.get)` in `get_cluster_rearrangement` depends on dict order, which isn't stable under python 2 -- worth a `sorted()` tiebreak in a file that otherwise pins `PGEN_SAMPLE_SEED` for reproducibility.
- **No tests.** Nothing covers `read_pgen_tables`, `_index_groups`, `_draw_conditioned`, `sample_pgen_median`, `rearrangement_lpgen` or the veto branch. I traced the numpy index arithmetic and it is correct, but it's correct-by-inspection only, and the two-line trivial case would be worth having: one true family of 10 sharing a rearrangement (should fire) against two families sharing one (should not).

#### What would settle it

1. Ablate the pgen clause -- tau = +inf, `n_shared >= 4` alone -- on d1/d3/d4. If the gain survives, the pgen machinery isn't the method, and "shared-rearrangement veto with a 4-member floor" is the simpler and more accurate description.
2. Sweep `PGEN_VETO_MIN_SHARED` over 2/3/4/6/10.
3. Report the 41 / 4 / 10 outcome-changed counts rather than `ctr['pgen_veto']`, and add `no_proposal` to the printed line so the categories sum.
4. Replicate d1 across seeds.
5. Fix and rerun the 50k validation so damage is measured, and replace the 0/9,149 with a test on data where the veto can fire.

None of this is an argument against the operator. A shared-rearrangement veto with a member floor is a reasonable thing to want, and the measured direction is right. I'd rather have it land described as what it does.

s-canchi (commented):
Thank you for the review.

1. **The k=4 threshold was swept, and the sweep was missing from the PR. Here it is**, re-run at the shipped cutoff, as change in correct and wrong pairs against the same partition with the guard off:

   | k | donor 1 | donor 2 | donor 3 | total |
   | --- | --- | --- | --- | --- |
   | 2 | +1,666 / +69 | +91 / +5 | +186 / +5 | +1,943 / +79 |
   | 3 | +1,290 / +24 | +64 / +0 | +145 / +0 | +1,499 / +24 |
   | 4 | +1,158 / +6 | +44 / +0 | +114 / +0 | +1,316 / +6 |
   | 6 | +1,032 / +0 | +32 / +0 | +92 / +0 | +1,156 / +0 |
   | 8 | +996 / +0 | +20 / +0 | +77 / +0 | +1,093 / +0 |
   | 10 | +961 / +0 | +0 / +0 | +0 / +0 | +961 / +0 |

   Note that k is a count of members sharing the rearrangement, not a probability threshold. It sets how much corroboration is required before rarity is allowed to matter at all.

   Wrong pairs fall steeply from k=2 to k=4 and are essentially gone by k=4: two of the three donors are already at zero there, and the third is at 6 against the 15,037 wrong pairs its partition already contains, which is 0.04 percent. Correct pairs fall monotonically throughout. k=4 is where those two curves meet: it is the smallest k at which damage has flattened out, and every step beyond it gives up correct pairs for almost nothing, 160 from 4 to 6 and a further 195 from 6 to 10.

   The reason I am comfortable with 4 rather than treating it as a tuned value is that it does not move. The same point holds across all three donors here, across three mutation levels, and across three independent simulation seeds, where the number of clusters it acts on holds to within a few percent; the per-seed counts are in point 5 below. A threshold fitted to one dataset would not behave that way. If you would still prefer exactly zero added wrong pairs, k=6 delivers that and the table above states what it costs.

2. **The rarity term was never ablated in the PR either. Ablated now, holding everything else fixed at k=4:**

   | run | correct pairs | wrong pairs |
   | --- | --- | --- |
   | shared-member condition alone | +2,517 | +1,497 |
   | shared-member and rarity together, as shipped | +1,158 | +6 |

   The rarity term removes 99.6 percent of the damage and keeps 46 percent of the benefit. Benefit to cost moves from 1.7 to 193.

3. **Outcome-changed counts, replacing the firing counts the PR quoted.** Clusters whose outcome actually changes are 41, 4 and 10 across the three donors, against firing counts of 1,537, 444 and 722. The firing counter overstates by 37x, 111x and 72x, because most clusters it counts would have been kept whole regardless.

4. **Effect on the three 200k donors:** +1,158 correct pairs for +6 wrong, +44 for 0, and +114 for 0. No regression in cluster purity, worst-cluster contamination, or the count of clusters holding 11 or more families.

5. **Replication across three simulation seeds at 30k**, since a small effect deserves the question of whether it is one draw:

   | sample | firings | correct pairs added | wrong pairs added |
   | --- | --- | --- | --- |
   | heavy, mutation 0.05 | 784 / 786 / 761 | +767 / +630 / +519 | 0 / 0 / 0 |
   | heavy, mutation 0.1 | 403 / 441 / 408 | +375 / +351 / +477 | 0 / 0 / 0 |

   Firing rates hold to within a few percent across seeds. Zero wrong pairs were added in any of the nine heavy samples, across three seeds and three mutation levels.

6. **Light chain is untouched, measured rather than asserted.** Zero firings in all nine light samples, with output byte identical every time, even though the largest number of sequences sharing a rearrangement reaches 1,335 there, well above anything on heavy. `refine_partition` returns after the light split and never reaches the operator this guard sits in.

s-canchi (commented):
1. **The 50k result is vacuous, as you say.** No rearrangement occurs even three times in those two samples, so the four-shared condition cannot be met, the guard never fires, and the partitions are byte identical with it on and off. I re-ran both end to end on regenerated data with the current pipeline to confirm rather than assume it.

   Worth adding, because it changes what the null means: those samples contain 49,975 and 49,983 distinct rearrangements in 50,000 sequences. Nearly every sequence is unique, so there is no convergence for a convergence-aware guard to respond to. That dataset cannot exercise this code path at all, so it is the wrong test rather than evidence of safety, and I should not have offered it as reassurance.

2. **The firing counter does overstate the effect**, as in point 3 of the previous comment. The split summary now also prints the `no_proposal` category, which was missing, so the reported categories sum.

3. **On the medians being close across donors.** They were measured separately, so their proximity does not by itself establish independence. The property the guard relies on is narrower: the cutoff is drawn from the parameter directory's own marginals, so it does not depend on how the input happened to be clustered. The docstring now claims only that, and the broader "generative property of the locus" wording is gone.

4. **On the comparison to #394, which I think rests on a misreading, so it is worth being precise.** Every generation-probability experiment in #394 was on the light loci. None of it was ever run on heavy. So this is not a failed approach being tried again. It reuses a concept that was evaluated before, on a locus where it was never evaluated, and to answer a different question.

   The difference in the question matters more than the difference in locus. In #394 the quantity was used to ask how many independent founders a cluster should be expected to contain, which is a counting question and needs the probability to be right in absolute terms. Here it is used only to ask whether the shared rearrangement is rare relative to the rest of the locus, which is an ordering question. Ordering survives the approximations in the factorisation; counting does not.

   Heavy is also where it should work better, and does: the higher naive diversity gives it far more distinct rearrangements to separate on than light has.

5. **On the suggested Poisson-style null specifically.** That form, comparing an expected number of independent founders against 1, is the counting question above. It was tried on light in #394 and fired on 83 to 97 percent of clusters, which is why it was not carried forward there. The rarity test here avoids it rather than repeats it.

   For whether the guard leaks into light at all, point 6 of the previous comment is the measurement rather than an argument from code structure: nine light samples, zero firings, byte identical output.

6. **The five code points are all real.** Magnitudes, so they can be weighed:

   | point | magnitude | resolution |
   | --- | --- | --- |
   | missing-key probability floor | 9, 1 and 2 firings depend on it | documented; it fails toward keeping a cluster whole, which is the conservative direction since the guard never creates a merge |
   | ties in the modal rearrangement | 13,009 ties, none change an outcome | sorted tiebreak added for determinism |
   | hand-copied table names | correct today, but unenforced | now derived from the existing column-dependency and filename helpers |
   | divide by zero in the sampler | unreachable on real files | guarded anyway |
   | the 0.25 insertion constant | cancels between sampler and scorer, flips 3 of 19,592 decisions | left unchanged, since changing it is churn |

7. **Following those up turned out one more case of the same kind.** A partly corrupt parameter directory silently switched the guard off and produced output byte identical to a deliberate off-run, with no way to tell the two apart afterwards. This one can corrupt a later result rather than only mislead a reader. It now fails loudly, and the check runs at the start of partitioning rather than inside refinement, so an incomplete directory stops the run in seconds instead of after vsearch and re-partitioning have finished.

8. **On tests.** You are right that there are none, and not only for this change. My practice has been scripted validation against fixed datasets before merging, which is real checking but is not a committed suite, and it should be one.

   I would treat this as a separate task rather than something to attach to this branch, not least because it is not obvious it belongs in this repository at all. I was thinking of discussing it on #337, or starting from any leads you have if you would rather point me at those.

s-canchi (commented):
The PR body is left as written so the review stays readable against it. Changes since:

1. The claim that passing 0 disables the guard is withdrawn. It was a mistake. The guard is not optional and there is no command-line switch for it, in keeping with the rest of refinement, which exposes nothing. The argument still exists internally for measuring with it on and off.
2. The k sweep and the rarity ablation are in comment 1. Neither was in the body.
3. Outcome-changed counts replace firing counts.
4. The 50k result is reframed. It confirms that the guard cannot fire without convergence; it is not a safety check.
5. Note on reading the gain: the subset of clusters where the guard actually changed an outcome accounts for +1,428 correct pairs on the first donor, against a net of +1,158. The correction runs in the opposite direction to the one the review assumed.

psathyrella:
>On tests. You are right that there are none, and not only for this change. My practice has been scripted validation against fixed datasets before merging, which is real checking but is not a committed suite, and it should be one.
>I would treat this as a separate task rather than something to attach to this branch, not least because it is not obvious it belongs in this repository at all. I was thinking of discussing it on https://github.com/psathyrella/partis/issues/337, or starting from any leads you have if you would rather point me at those.

This all sounds reasonable, thanks.

>If you would still prefer exactly zero added wrong pairs, k=6 delivers that and the table above states what it costs.

No, it makes sense in general to balance over- and undermerging. 

edit: `data/mebcell-mutual-info.jpg` is a better source of correlation info than CC's garbled comment mentioned below, although they're both from the same place in the code.

CC below

psathyrella:
> Claude Code:

The ablation settles the mechanism question against my review, so let me say that plainly first: the rarity term is load-bearing, and I was wrong to call it decoration. +2,517/+1,497 without it against +1,158/+6 with it is the number that matters. My error was reading the 50 percent admission rate as low selectivity -- how many rearrangements pass is not the same as which ones, and the half it rejects is where nearly all the damage lives. A filter that turns away half of everything can still remove 99.6 percent of the errors if the errors concentrate in that half, which is what your table shows.

k=4 is fine as shipped. The sweep gives the whole trade curve, so it is a choice rather than a correctness question, and +6 wrong pairs against the 15,037 the partition already carries is a good price for 160 correct ones. I would rather accept some over-merging than demand zero of it -- an over-split and an over-merge are both errors and there is no reason to weight one at infinity. If you prefer k=6 on the lineage-inference argument from #394, that is also defensible; I do not need it changed.

I checked the new commit. The derived specs produce byte-identical filenames to the hand-written ones, the sorted tiebreak and the `probs.sum() <= 0` guard are right, `validate_pgen_tables` running up front is a better fix than the one I asked for, and the `no_proposal` category now makes the printed counts sum. The 50k reframing, the outcome-changed counts, and the narrowed cutoff docstring all land.

Your #394 reading is better than mine. Counting needs the probability right in absolute terms and ordering does not, and every one of those experiments was on light. I read a locus-and-question-specific result as a blanket one.

#### Code reuse: one of three

`utils.column_dependencies` and `get_parameter_fname` are now the source of truth, which is the fix I wanted. The other two are untouched, and the first is the one I care about most:

1. **`recombinator.read_vdj_version_freqs` (`recombinator.py:241`) and `choose_vdj_combo` (`:469`) already draw complete rearrangements from a parameter dir**, using the joint `all-probs.csv` rather than a product of marginals. `sample_pgen_median` is a second implementation of that same operation. Two things follow, and they are separate:

   - Duplication on its own terms. There are now two pieces of code in partis that answer "draw a rearrangement from this parameter dir", with different dependency structures, and nothing keeps them consistent. This is the same failure mode as the hand-copied filenames, one level up.
   - It is also the model the simulated data was made with, which is the subject of the next section.

2. **`recombinator.read_insertion_content` (`recombinator.py:102`)** reads the real per-base insertion content and falls back to uniform only when the data is absent. You answered the constant -- it cancels between sampler and scorer and flips 3 of 19,592 decisions -- which I accept as a reason not to change the number. It is not a reason to keep a second, worse copy of a thing partis already reads correctly. Same objection as (1), smaller stakes.

I would rather these were addressed as reuse than as accuracy, because that is where the cost is: the accuracy effects are small and you have measured them, but each duplicate is a place where the two copies can drift apart silently, and this module has now produced that exact bug twice.

#### How much does the independence assumption cost? This needs measuring

This is the one I would most like a number on, and it is a question about accuracy rather than about code.

`rearrangement_lpgen` scores a rearrangement as `P(V) x P(D) x P(J)` times deletion and insertion-length terms conditioned only on their own gene. That treats the segment choices as independent. They are not: there are substantial correlations among these parameters, and partis says so itself in `utils.column_dependencies`, where the `v_gene` entry is commented "v choice actually depends on everything... but not super strongly, so a.t.m. I ignore it". Ignoring them is a defensible modelling choice for the HMM, which is why your factorisation matches it. It is a different matter to build a rarity test on top of it, because a product of marginals systematically under-weights exactly the combinations that correlation makes common. Those rearrangements read as rarer than they are, and rarer is the direction that fires the guard.

The simulated data sharpens this rather than softening it. `recombinator` built the d1/d3/d4 sequences by sampling whole rows from `all-probs.csv`, which carries those correlations; the veto then scored the same rearrangements as a product of marginals. So the guard may be firing on rearrangements it treats as rare that the simulator was in fact producing frequently, and every number in this PR was measured under that mismatch.

Your ordering argument is a good reason this need not be fatal, and I think it is probably right -- ranking is far more robust to the approximation than absolute values are. But it is currently an argument that the approximation survives rather than a measurement that it does, and the whole test rests on it. Three things would settle it, cheaply, on one donor:

1. For every rearrangement observed in `all-probs.csv`, compare `log10 P(r)` from the factorisation against the empirical joint frequency. A scatter and a rank correlation say directly how much ordering the approximation costs. If the rank correlation is high, your argument is measured rather than asserted, and the PR is stronger for it.
2. Report the size of the bias and which way it runs -- how much the factorisation under-weights the most correlated combinations, since that is the population that could be wrongly held.
3. Recompute the cutoff and the held set using the joint, and check whether the same clusters are held. If the outcome is unchanged, the approximation is settled for this use and can be documented as such.

If it turns out the ordering does not survive, the joint is already read by `recombinator` and would be the fix rather than a rewrite.

#### The naming point went unanswered in prose, and half-answered in code

The docstrings now say `P(r)` and `rearrangement_lpgen` carries "comparable only against values from the same factorisation", which is the substance of what I asked for. But all 55 identifiers still read `pgen` -- `read_pgen_tables`, `sample_pgen_median`, `PGEN_VETO_MIN_SHARED`, the printed log line -- and nothing states that this is not the published quantity. `P_gen` in Murugan 2012 and OLGA (Sethna 2019) is a marginal over every scenario consistent with an observed sequence; this is `P(r)` for the one scenario partis inferred, under an independence-factorised model fit by counting hard calls. A reader of the docstrings is now correctly informed and a reader of the symbols is not. Renaming to something like `rearr_prob` or a rarity score would close it, or one sentence in the module header if you would rather not churn the names.

#### Unanswered: no fraction-of-cluster condition

`n_shared` is compared to an absolute 4 with no requirement that those 4 be any particular share of the cluster, and uids with incomplete features are dropped from the count silently. The k sweep answers how much corroboration is needed; it does not answer "4 out of how many". As written, 4 members sharing a rearrangement holds a cluster of 400 whole. This module already has a rearrangement-*majority* guard, so the asymmetry is worth either closing or justifying.

The concrete question I asked and did not get: **what is the largest cluster the guard held on d1?** If the answer is that vetoed clusters are all small, this is a non-issue and worth one line in the PR. Your point that the guard never creates a merge is true but does not bound this, since preventing a split of a large mixed cluster leaves it mixed.

Two smaller ones: `PGEN_MIN_PROB = 1e-9` now says what it does, but not why 1e-9 rather than 1e-6 or 1e-12 -- if it was picked as smaller than the smallest observed table entry, say so. And the 0/9,149 figure is moot now that the claim is withdrawn, so no need to chase where it came from.

#### Tests

To answer your question directly: [loris-experiments/tests](https://github.com/matsengrp/loris-experiments/tree/main/tests) is the model I had in mind, and yes, it belongs in partis rather than elsewhere -- the code under test lives here, and a suite that ships separately from the code it checks will rot. Partis has correctness-style tests and could use more of the detailed agent-style kind alongside them.

#337 works for the discussion. Scoping it as its own task rather than attaching it to this branch is right.

Worth being concrete about why this is not bookkeeping. Of the two real bugs found on this branch, one was a silently corrupt parameter dir producing output byte-identical to a deliberate off-run, and the other was a fail-open path that turned the guard off without saying so. Both are invisible to end-to-end validation against fixed datasets, because both produce output that looks fine. That is the class of bug a unit suite catches and scripted validation structurally cannot.

s-canchi (commented):
Thank you both. On the ablation, agreed.

1. Independence check, on all three donors. I compared the factorised score against the true joint distribution (`all-probs.csv`, the exact distribution the simulator draws from, not an estimate), then recomputed the held set with the joint score in place of the factorised one.

   | donor | rho(joint, factorised) | mean bias | factorised cutoff | joint cutoff |
   |---|---|---|---|---|
   | 1 | 0.07 | +5.8 | -16.26 | -5.30 |
   | 2 | 0.11 | +5.7 | -16.64 | -5.17 |
   | 3 | 0.08 | +5.8 | -16.58 | -5.19 |

   | donor | shipped held: impure / wrong pairs | joint's extra holds: impure / wrong pairs |
   |---|---|---|
   | 1 | 0.5% / 66 | 22.5% / 7,710 |
   | 2 | 0% / 0 | 17.2% / 844 |
   | 3 | 0% / 0 | 13.6% / 1,040 |

   No correlation on any donor, and the joint veto is a strict superset of the shipped one on all three (it releases nothing the shipped veto holds) while being far less selective. Switching to the joint would be a regression on every donor, not a fix.

   I do not have an explanation for why the factorised score is this selective if it is not measuring rarity as claimed. I would rather say that than guess.

2. Naming. All 55 `pgen`-bearing identifiers renamed to `rearr_prob`, including the printed log line and docstrings.

3. Largest cluster held. 150 members, donor 1, single true family. No impure held cluster on any donor is larger than the 56-member case already flagged, and that one is also pure.

4. `PGEN_MIN_PROB`, now `REARR_PROB_MIN`. This is a defensive floor, the same role `utils.eps` plays elsewhere in the codebase: chosen to sit safely below anything a real parameter dir could produce, not tuned to a specific dataset. Measured on donor 1: the smallest nonzero probability actually in the tables is about 1e-5, so the floor sits about four orders of magnitude below the smallest real signal, comfortably clear of it rather than close enough to compete with a genuinely rare observation.

5. Code reuse, one of two closed. `sample_rearr_prob_median` stays on the factorised draw rather than `recombinator`'s joint reader, and point 1 is why. The insertion-content duplication is fixed: both functions now read the real per-base composition from the same `<boundary>_insertion_content.csv` `read_insertion_content` reads, falling back to uniform only when that file is absent. Effect on donor 1: cutoff -16.26 to -16.12, vetoed count 1,537 to 1,534.

psathyrella:
> Claude Code:

Points 2 through 5 are closed from my side: the rename, the largest-held-cluster answer (150 members, single true family, and no impure hold above the 56-member case), the floor justified against the 1e-5 smallest real entry, and the insertion-content reuse. Declining the joint reader is now a measured decision rather than an unexamined one, which is what I wanted.

Thank you for running the independence check and for reporting the answer straight rather than fitting a story to it. I think the explanation is available, and it is a good one for the operator.

#### Why the factorised score is selective

Measured on donor 1, over the 154,079 distinct rearrangements in `all-probs.csv`:

- **The factorised score is almost entirely total insertion length.** Spearman correlation between `log10 P(r)` and `len_vd + len_dj` is **-0.976**, and **79.5 percent** of the score's variance comes from the `0.25^L` term by itself. Longer junction, lower score, "rarer".
- **The joint has almost no resolution to correlate with.** **88.7 percent** of distinct rearrangements appear exactly once, and the median count is 1. The empirical joint frequency is therefore a near-constant 1-versus-1 tie across most of the space. That is also why its cutoff lands at -5.30, which is log10(1/200,000): "appeared once".

Both halves of the puzzle follow. The two scores do not correlate because one of them barely varies -- rho of 0.08 is what you get comparing a smooth continuous quantity against a mostly-flat one. And the factorised score is selective because it is measuring **junction length**, not rarity relative to the repertoire.

That also explains why the joint is a strict superset that holds far more: with most counts at 1, thresholding it admits nearly everything shared by 4 members, including the convergent cases, which is exactly the 22.5 percent impure population you measured.

#### This is a better justification than the one in the PR

Long N-insertions are close to the ideal evidence for the question the guard asks. Convergent recombination reproducing a long random junction independently is very unlikely, so "4 or more members share a rearrangement with a long junction" is strong evidence of common descent, and it is strong for a mechanical reason rather than a statistical one. The operator is doing something sound. Its stated mechanism has now been wrong twice, and the real one is simpler and easier to defend than either.

So the follow-up worth running is whether the rest of the machinery earns its place: does a plain total-insertion-length threshold, at the same `k = 4`, reproduce the held set and the pair counts? If it does, the tables, the million-draw sampling, the parameter-dir dependency, `validate_rearr_prob_tables` and every fail-open path we have been repairing all come out, and what remains is two integers per cluster. The other 20 percent of the score's variance, from the gene and deletion terms, is what would show up as a difference, and that is precisely what the comparison measures.

If length alone does not reproduce it, that is also worth knowing, because then the gene and deletion terms are contributing something real and the current form is justified on evidence.

#### One correction of my own

I said earlier that the factorisation "matches partis's own HMM factorisation", pointing at `utils.column_dependencies`. That was wrong, and it was the strongest form of the "it is already partis's model" defence, so it should not stand. `column_dependencies` is the parameter-file counting schema: it sets which columns `parametercounter` writes and what each is conditioned on (`parametercounter.py:32,230`), and `hmmwriter` reads erosion and insertion probabilities back under that same conditioning (`hmmwriter.py:424,505`). It says nothing about gene choice -- `v_gene`, `d_gene` and `j_gene` have empty dependency lists, meaning only that each is counted on its own -- and `recombinator`, which is partis's actual generative model of a rearrangement, does not use it at all. Matching that schema is not the same as matching a partis model of how likely a rearrangement is. Your measurement has overtaken the question either way.

s-canchi (commented):
Thank you for the analysis. Confirmed by measurement: a plain `len_vd + len_dj` threshold at the same `k=4` reproduces the veto's effect. The length cutoff comes from the parameter dir alone: draw a million rearrangements from its own d and j gene and insertion-length distributions, then take the median of len_vd + len_dj.

Cluster level first:

| donor | candidates (n_shared>=4) | fact_vetoed | length_vetoed | agree | length cutoff |
|---|---|---|---|---|---|
| d1 | 3,480 | 1,534 | 1,646 | 3,224 (92.6%) | 11.0 |
| d3 | 996 | 445 | 455 | 940 (94.4%) | 12.0 |
| d4 | 1,571 | 720 | 728 | 1,487 (94.7%) | 12.0 |

Pair level:

| donor | veto-OFF correct/wrong | factorised ON | length-only ON |
|---|---|---|---|
| d1 | 558,537 / 15,037 | +1,158 / +6 | +1,156 / +0 |
| d3 | 43,829 / 2,114 | +44 / +0 | +44 / +0 |
| d4 | 69,131 / 2,298 | +114 / +0 | +114 / +0 |

d3 and d4 are outcome-identical between the two scorers. d1 is strictly better with length-only: same correct-pair gain, zero of the six wrong pairs the factorised score's extra holds introduce. The gene and deletion terms are not earning their place on any of the three donors.

The code is updated with this refactor (`ce0d9db84`).

I also looked at which cdr3-length groups drive the 200k gain, since that is the axis the veto reasons over. d1's gain is spread across many groups, with no single one dominating. d3's gain comes mostly from one group (cdr3-51). d4's gain is spread across a handful of groups. For the 1M check below, I picked one cdr3 group per donor from this breakdown, each a representative top contributor for that donor.

The results also hold at 1M scale:

| donor (cdr3 group) | cutoff | correct/wrong delta |
|---|---|---|
| d1 (cdr3-30) | 11.0 | +0/+0 |
| d3 (cdr3-51) | 12.0 | +814/+0 |
| d4 (cdr3-75) | 12.0 | +0/+0 |

psathyrella (approved):

Issue #396, opened 2026-07-26
Update germlines for rhesus macaques


ksung25:
There are some rhesus macaque BCR data we would like to process as part of [this data-central issue](https://github.com/matsengrp/data-central/issues/135).
There are recently updated germline sets from OGRDB [here](https://ogrdb.airr-community.org/germline_sets/Macaca%20mulatta). Is it worth updating to these?

psathyrella:
Yes! My impression is that germline sets are still so hopelessly incomplete that it's worth updating with any new high quality genes. (They do need to be high quality, since adding spurious genes is worse than having missing ones).

psathyrella:
I've had on my todo list for ages to add the engelbrecht/rodriguez/etc germlines as well, probably from the merged set in vanwinkle-170/, so feel free to do that while you're at it (or lmk if I should do all of em).

ksung25:
Could you add the germlines, please? For one thing, I don't seem to have permission to even upload a branch to the remote repo, and another thing is I'm not sure how or what  are the engelbrecht/rodriguez germlines to add.

psathyrella:
Should be all set, let me know if you still have trouble. Workflows not quite yet in pypi but will be soon 

https://github.com/psathyrella/partis/actions
https://pypi.org/project/partis-bcr/

PR #401, opened 2026-09-01
Settle hfrac bin-size and test refine at scale

s-canchi:
Main goal: test and resettle the hfrac-based bin size cap (`HFRAC_MIN_SEQS_DEFAULT` / `HFRAC_MAX_BIN_SIZE_DEFAULT`). Default threshold raised after testing at 1M scale, from 50000/100000 to 240000/200000. Performance, correctness, and composition data in the comments below.

Two further changes fell out of that work:

- Light-chain refine now derives its shared-descent split cutoff per bin from a fitted null tail, replacing a fixed constant. Details in the light-chain comment.
- Merging hmm parameter tables across cache-parameters chunks now sums per-position mutation counts across chunks instead of leaving a gap when chunks disagree.

Also included: unified standalone and integrated disjoint-groups actions onto the same `--paired-outdir` layout with unconditional naive-vsearch, so both code paths produce identical output structure.

Figures and numbers below reflect runs against the updated germline set.

This is the first PR that runs refine itself end to end on a full 1M set. 013's length veto ran refine at 1M, but only on select subsets as a proof point, not the full set. Full standard-partition/subset-partition runs at 1M+ were done separately, on the real validation set (soto-bcr), before this feature's dev work started, but those used vsearch/HA/subset-partition only, never refine, never disjoint grouping. So this dataset checks two things: does refine hold up at that combination and scale (composition, correctness), and is the pipeline tuned for memory/runtime at that scale (the hfrac-based bin size cap work).

#### Pipeline stages, referenced below

```
raw seqs, one same-CDR3-length group
        |
        v
  [ vsearch ]        naive clustering, cheap, coarse
        |
        v
  [ HA-repartition ]  hmm-based re-annotate + subdivide
        |
        v
  [ refine (final) ]  shared-descent split / naive-identity merge
        |
        v
  final partition  ---scored against--->  truth (ground-truth simulated partition)
```

Every "vsearch vs HA vs refine vs truth" figure below is scoring each stage's own output
partition against the same ground truth, not against each other's runtime. Refine's input is
HA's output, which is itself shaped by vsearch's output, so refine's numbers should be read
next to both prior stages, not in isolation.

---

**Runtime, memory, and the hfrac-based bin size cap**

#### Runtime and memory: vsearch vs refine per job (HA shown separately), 1M scale

fig_perf_scaling

- Each point is one per-task (vsearch or refine) job across all 9 donor/locus groups at 1M, full raw distribution not just the max. Both stages grow with group size, refine stays below vsearch everywhere measured, all points at or under the 200000 cap.
- HA excluded from the scatter (bundled, not comparable per-group), shown as a per-locus box distribution instead.

200k peaks, largest group per locus, all well under the 1M numbers above:

| Locus | Group | n_seqs | vsearch mem (GB) / time (m:s) | refine mem (GB) / time (m:s) |
|---|---|---|---|---|
| igk | cdr3-33 | 89,366 | 2.09 GB / 3:54 | 1.87 GB / 2:09 |
| igl | - | - | 0.93 GB / 3:47 | 0.91 GB / 0:55 |
| igh | - | - | - | 0.72 GB / 0:18 |

#### Why raise the hfrac-based bin size cap to 240000 / 200000, and does it cost anything?

- Refine is single-threaded per job. Max bin size is capped to per-job memory/wall-time, allowing many jobs to run concurrently. Tested and validated on all loci at 1M.
- igh refine cost is O(U^2), U approx 0.81 * n_seqs, stable within about 2% across 3 measured points. Unsplit 114,188-seq group costs 3.24x the split runtime for under 0.003 difference in purity/completeness.
- Entry gate = cap * bin-pack tolerance removes split-then-reglue waste: d1 1M hfrac-firing groups drop 15 to 1 (igk cdr3-33 only), final bin count roughly unchanged (149 to 142).
- Final validation set (real data) has a single igh CDR3 group of about 2 million sequences per donor. Cap needs to hold at that scale too, not only on current simulated sets.

#### Does raising the hfrac-based bin size cap change correctness?

No, the split/reglue mechanism is correct at any cap value; raising it only removes waste.

- Old cap (50k/100k): igh group of 114,188 sequences split into 50,084/64,104. 0/47,575 families crossed the boundary. Whole vs split purity 0.9326/0.9350, retention differs by under 0.003. Same group under the germline update: still cdr3-48, size unchanged within noise -- biggest CDR3-length group per donor/locus is unchanged, cap-selection numbers still hold.
- Light (igk cdr3-33, 305,992 families): 13 families crossed the boundary at the old cap, drops to 3 at the new cap (close V-gene paralogs, e.g. `IGKV2-40*01` len 340 vs `IGKV2D-28*01`/`-29*01` len 337, lose naive registration and split across the hfrac gap). Pre-existing mechanism, cap-dependent not a threshold defect: same 3 families split at old germline too.
- Net: subdivision mechanism is correct independent of cap value. Raising the cap removes wasted compute and also reduces the residual paralog-driven crossings on light.

---

**Contamination and composition, 1M sets**

#### Metric definitions

`contam_seq_frac`: contamination as a fraction of sequences across all inferred clusters.
`mean_contam`: mean contamination per inferred cluster (fraction outside the plurality family), per bin per stage. 
`mean_reten`: mean fraction of a true family captured by its single best-matching inferred cluster, per bin per stage. 
Truth-vs-inferred figures: dashed line is `n_true_families`, solid/marker lines are `n_inferred_clusters` per stage, both per bin. Bins group true families by size: 2-4, 5-19, 20+.

#### Why does the inferred line sit below the true line at every bin? (raw counts, not normalized)

fig_cluster_count_vs_truth_igh
fig_cluster_count_vs_truth_igk
fig_cluster_count_vs_truth_igl

- Each bin is a raw count of clusters/families of that size, not a fraction of a fixed total. Total inferred cluster count is not constrained to equal total true family count, so the inferred line sitting below the true line at every bin does not require it to be above somewhere else.
- Usually means: many small true families were merged into fewer, larger inferred clusters, so total inferred cluster count is lower across the board. Not evidence of mass shifting from one bin to another.

#### Does the recovered cluster-size shape match truth, and does that speak to composition?

fig_size_cdf_by_stage_1m_igh
fig_size_cdf_by_stage_1m_igk
fig_size_cdf_by_stage_1m_igl

Light: yes, and dramatically, refine reshapes giant pre-refine blobs (vsearch/HA tail past 10^4) almost onto truth. igh: shape never moves, at any stage, because it was already right pre-refine. That is a shape statement only, not a composition one: igh still has a real, scale-growing contamination gap (below) even though its shape never changes.

Same pattern holds for 200k as well. 

#### Does contamination get worse at 1M vs 200k?

fig_contam_reten_by_bin_200k_igh
fig_contam_reten_by_bin_200k_igk
fig_contam_reten_by_bin_200k_igl

Yes for igh, no for light.

- Both scales scored against final `partition-refine-<locus>.yaml` output, directly comparable.
- igh locus-wide `contam_seq_frac`: 0.050 (200k) to 0.103 (1M), 2.1x.
- Light: final mean contamination fell at 1M vs 200k (0.021-0.287 vs 0.042-0.493 range), opposite direction from igh.

#### Does refine improve composition per bin, without costing retention?

fig_contam_reten_by_bin_1m_igh
fig_contam_reten_by_bin_1m_igk
fig_contam_reten_by_bin_1m_igl

- **Heavy (igh): yes.** Contamination drops by half to two-thirds in the 2-4 and 5-19 bins at every donor (e.g. d1: 0.23 to 0.07, 0.23 to 0.17, HA to refine). The 20+ bin is already low (3-10%) and stays there, within a point or two of noise. Retention stays flat to improved in every bin, refine does not trade completeness for purity here.
- **Light (igk/igl): yes on contamination, no on retention.** Contamination drops sharply in every bin (e.g. d1-igk 2-4: 0.41 to 0.01), but retention collapses in every bin too, not only the overmerged 20+ bin (d1-igk: 2-4 0.98 to 0.50, 5-19 0.98 to 0.22, 20+ 0.98 to 0.09). This is the recovery ceiling (light-chain section below): shared-descent splitting only keeps members with a detectable shared mutation, and that shortfall hits small true families as much as large ones, not just the big overmerged clusters.
- igh's 20+ bin has lower vsearch/HA contamination than its own 2-4/5-19 bins (0.03-0.12 vs 0.23-0.39): heavy clusters are not more contaminated just for being bigger.
- HA's own effect (vsearch to HA, before refine) is small everywhere and does not help large bins more than small ones.

#### Where the contamination comes from: vsearch, HA, or refine

fig_stage_progression

- Same `mean_contam` metric as the bin figures, aggregated per donor/locus instead of per bin (1M only).
- HA barely moves anything.
- igh: refine roughly halves contamination (e.g. d1 0.24 to 0.10); the 200k-to-1M rise above is mostly inherited from vsearch.
- This aggregate is cluster-count-weighted, dominated by the numerous 2-4 bin: d4-igl's aggregate (0.10) looks solved, but its 20+ bin alone is still 0.74 after refine (1,034 vs 8 clusters, per the bin figures above).

#### Anomaly

d3 runs hotter than d4 on igh at both scales (5-19 bin: 0.247 vs 0.199 at 1M), concentrated in about 10 shared largest CDR3-length groups, not a broad donor effect. Same d3-vs-d4 crowding flagged on light in issue #337; not attributable from simulation alone.

---

**Light-chain refine**

#### What changed

- False link rate: light-chain refine can only ever split, never merge (no merge signal on light). To split, it looks for pairs of sequences inside one cluster that share a mutation pattern only expected from true clonal descent, and uses that as evidence to keep them together. "False link rate" is the fraction of those accepted links that actually connect two different true families, not one -- lower is better.
- Old fixed constant (`WEIGHTED_DESCENT_ALPHA = 0.005`, one global p-value cutoff applied to every bin regardless of that bin's own noise): measured false-link rate 72.4-96.3% across all donors, both light loci, both scales.
- `derive_bin_alpha` replaces that single global cutoff: fits a cutoff per bin from that bin's own null p-value tail instead.

#### Failure mode and design

- Composition-overlap failure: fit succeeds, but null/signal overlap too much for `fdr:0.05`.
- Design: primary `fdr:0.05`, retry `fdr:0.25` on failure (resolves 7/10 overlap bins). If both fail, no action is taken and refine refuses to split that bin. Purity-first: the 3 bins that never resolve keep poor baseline purity (0.03-0.23) rather than force a threshold.

#### Recovery ceiling

- Retention can never reach 100%: shared-descent splitting only recovers members with a detectable shared mutation. The rest are structurally unrecoverable regardless of cutoff.
- Ceiling = fraction of family members sharing detectable mutations, measured directly at 1M: grand total 0.377 (233,648/620,295 members, both light loci, all donors). Splits by donor, not just aggregate: d1 0.47-0.53, d3/d4 0.13-0.18 (both loci).

#### Scale invariance, 200k vs 1M

| Group | Scale | n_seqs | Cutoff |
|---|---|---|---|
| d3-igk cdr3-36 | 200k | 21,061 | 3.16e-20 |
| d3-igk cdr3-36 | 1M | 513,241 | 3.16e-20 |
| d3-igl cdr3-36 | 200k | 14,508 | 3.16e-18 |
| d3-igl cdr3-36 | 1M | 280,224 | 1e-18 |
| d4-igl cdr3-33 | 200k | 67,462 | 1e-17 |
| d4-igl cdr3-33 | 1M | 1,808,837 | 3.16e-18 |
| d1-igl cdr3-39 | 200k | 2,840,963 | 3.16e-19 |
| d1-igl cdr3-39 | 1M | 76,003,090 | 3.16e-12 |

All 8 measurements pick the same tier (`fdr:0.05`) at both scales. Discrete tier is scale-invariant; the one continuous fit shift (d1-igl cdr3-39) is orthogonal to it.

#### Known gap

Per-uid p-value cache in the shared-descent splitter scales with each uid's mutation count (SHM), not just pair/sequence count. Leading suspect for the 8.65 GB peak on the worst bin (d1-igk cdr3-33). Current sets have low, capped SHM, so this has not been stress-tested.

---

**Heavy-chain refine**

#### Representative CDR3 groups, 200k to 1M

| Group | max_cluster 200k | max_cluster 1M | contam_seq_frac 200k | contam_seq_frac 1M |
|---|---|---|---|---|
| d1 cdr3-48 | 460 | 217 (shrinks) | 0.100 | 0.235 |
| d3 cdr3-63 | 36 | 211 (5.86x) | - | roughly doubles |
| d3 cdr3-51 | 51 | 95 (1.86x) | - | - |
| d4 cdr3-63 | 30 | 147 (4.90x) | - | - |
| d4 cdr3-51 | 48 | 185 (3.85x) | - | - |

Contamination rises 200k to 1M in every group regardless of cluster-size direction (d1 cdr3-48: cluster shrinks, contamination still more than doubles). Cluster size tracking group size on heavy (unlike light, which is flat) does not explain this: within scale, bigger clusters are not more contaminated (near-zero correlation). Better explained by naive crowding: heavy's distinct-naive count scales linearly with group size (no saturation), so at 1M more distinct families' naive sequences sit within hamming range of any given cluster, independent of whether that cluster itself grew.

#### Germline-update outcome on igh

- Checked the biggest CDR3-length group per donor/locus, 1M, old vs new germline: same CDR3 length and size within a handful of sequences everywhere -- table above unaffected.
- Ran A/B against germline update: igh unchanged, every field within 1 percentage point, both scales.
- Largest move: d1-igh 1M mean contam 0.105->0.096, purity 0.925->0.946, both cleaner, within noise.
- d1-igh's singleton overshoot (4.4%->5.6% at 1M, same-commit rerun) has a direct germline mechanism: same V/D/J calls and reference sequences old vs new, but `n_mutations` per uid shifts by ~1 anyway (new alleles move the HMM's erosion/insertion priors), tipping refine's merge gate on borderline splits. Most visible on d1 since it is the highest-SHM donor, not a d1-specific bug.

---

**Germline update**

#### What changed

- Cherry-picked germline update: `d0e4f57b2` (default germlines: engelbrecht/rodriguez human + newer OGRDB macaque, #396/#398), `c4fc1a55f` (docs-only follow-up).
- Ran A/B against the update, full pipeline, no other code change: no measurable regression in composition, contamination, purity, completeness, or per-task runtime/memory, at either scale.

#### `subset-partition` crash

- Local test suite (`partis-test.py --bust-cache`): 3/4 combos clean; `--slow --paired` crashed in `re_pad_hmm_seqs` (`partis/utils.py:4761`).
- One cluster has members with a J-end pad of 82 vs 81 -- ordinary indel/J-deletion variance, consistent V-gene assignment, germline-invariant, not a data error.
- Not germline-update-specific: `--disjoint-groups` under `subset-partition` is absent on main, so this test scenario only exists on the disjoint-grouping feature branch.
- Test-only: `re_pad_hmm_seqs` is reached only via `--continue-from-input-partition` (`subset_partition`'s merged-run path); zero hits elsewhere in the refine/cluster pipeline.
- Side effect: the crash killed the whole `--slow --paired` bust-cache combo before it wrote any
  output, so all of `test/paired/ref-results-slow/`  is still the originally inherited data, not regenerated against this branch's own code.
  
**Proposed fix** (not applied yet):

```python
# current -- crashes on legitimate per-sequence variance:
sw_ldist, sw_rdist = [get_single_entry(list(set(l))) for l in [sw_ldists, sw_rdists]]

# proposed -- take the max padding needed across the cluster, not "the one value":
sw_ldist, sw_rdist = [max(l) for l in [sw_ldists, sw_rdists]]
```

Safe: pad is applied as flanking N's, so a member needing less than the cluster max just gets a few harmless extra Ns.


psathyrella:
I'm not sure I understand the big picture here. It sounds like the goal is testing on larger-than-before samples, but I thought you'd run on samples with 1M sequences a lot before?

What does "performance axis" mean? It sounds like there's some other axis, but I don't see one, nor do I see a plot with performance on an axis.

"runtime and memory by stage": I would think 'stage' would mean different stages of the refinement pipeline, but neither upper vs lower plots, nor left vs right seem to be stages. Maybe the green/blue colors? But then how does HA fit in? I think basically, what message am I supposed to take away from this section? It feels like CC is telling me lots of things, but I'm not sure why it's telling me each of them.

`Why 240000 / 200000` 
`Correctness, before and after the cap change`

I don't understand either of these, but maybe they're more for internal use.

`Truth vs inf clusters by stage`

I would have expected that when comparing family size distributions, if a distribution is higher than another on (e.g.) the right, that it'd also have to be lower on the left. Maybe I'm just expecting normalization on the y axis, and I'm not saying it has to be normalized, but I can't figure out how to interpret the green line that's uniformly vastly lower than the true line. There's way fewer singletons than in the true partition, and they were absorbed into a relatively small number of large families (no, that doesn't make sense, then the high end of green would be higher than black).

 I really can't make sense of all the plots below this. Partly there's just so many and I can't figure out the differences among each, or what all different samples are run on. What is the message of all of this? What is our expectation? What are we hoping to learn?

`Light-chain refine`

I unfortunately will also need explanation here as well, for instance:

>Old fixed constant (WEIGHTED_DESCENT_ALPHA = 0.005): measured false-link rate 72.4-96.3% across all donors, both light loci, both scales.

what is false link rate? Are those values good or bad? what is scale -- is it just always a synonym for size now?

>derive_bin_alpha fits a cutoff per bin from that bin's own null p-value tail instead.

this is in the 'changed' section, but I can't figure out what this is saying changed -- e.g. instead of what?

I have similar questions with the rest. I think the basic issue is that CC doesn't understand how to explain things to people with different contexts -- it thinks everyone knows everything that it knows, whereas I only understand the broad strokes of what you're doing, but have no idea of the details that are changing from version to version.

s-canchi:
Additional open question: could we include an option to keep the final partition output as separate files instead of merging into one single file? The size of the final `*.yaml` file tracks with sequence count. The problem is not only on the file size growing, but one large file increases chances of a crash leading to truncated corrupted file; multi part files also make data access and debugging easier. 

If we want to continue the single file format, then can we switch to more optimal format (e.g. Parquet)?

psathyrella:
> Additional open question: could we include an option to keep the final partition output as separate files instead of merging into one single file? The size of the final `*.yaml` file tracks with sequence count. The problem is not only on the file size growing, but one large file increases chances of a crash leading to truncated corrupted file; multi part files also make data access and debugging easier.
> 
> If we want to continue the single file format, then can we switch to more optimal format (e.g. Parquet)?

Either seems defensible to me, but I guess keeping the same format seems simpler. I don't see a problem with, for large samples where we're using this whole new workflow, keeping multiple output files.

s-canchi (commented):
#### Pipeline stages, defined once

The word "stage" gets used loosely below. One picture, defined once, instead of re-explaining it per section:

```
raw seqs, one same-CDR3-length group
        |
        v
  [ vsearch ]        naive clustering, cheap, coarse
        |
        v
  [ HA-repartition ]  hmm-based re-annotate + subdivide
        |
        v
  [ refine (final) ]  shared-descent split / naive-identity merge
        |
        v
  final partition  ---scored against--->  truth (ground-truth simulated partition)
```

- Every figure that scores "vsearch vs HA vs refine vs truth" is scoring each stage's
  own output partition against the same ground truth, not against each other's runtime.
- Refine's input is HA's output, which is itself shaped by vsearch's output, so refine's
  result is always read next to both prior stages, not in isolation: a change in refine's
  numbers can come from refine itself, or from what it was handed.
- The runtime/memory scatter only covers vsearch and refine as per-job points (HA is
  bundled, shown as a separate box distribution), that is a compute-cost comparison, a
  different question from the composition/correctness figures above.

#### Big picture

- Two separate 1M-scale things happened before this PR, neither is what this PR is doing.
- 013's length veto did run refine at 1M, but only on select subsets drawn from the 1M set, as a proof point, not the full dataset end to end.
- Full standard-partition and subset-partition runs at 1M+ scale were done separately, on the real validation set (soto-bcr), before any of this feature's dev work started. Those used vsearch/HA/subset-partition only, never refine, and never disjoint grouping.
- This PR is the first time disjoint grouping plus refine has been run together on a full 1M set end to end.
- So the purpose of this dataset and these runs is two things: check refine holds up at that combination and scale (composition, correctness), and optimize the pipeline for memory/runtime at that scale (the hfrac-based bin size cap work).

#### Does raising a job's per-group data size cost runtime or memory? ("performance axis" / "runtime and memory by stage")

No, both stay bounded, that is the point of the hfrac-based bin size cap; here is the evidence and what the labels mean.

- "Performance axis" just means runtime and memory, as distinct from the correctness and composition sections that follow it. Bad label, will fix the heading.
- "By stage" is the vsearch/HA/refine picture above. The scatter itself is only vsearch vs refine, one point per job, directly comparable per group; HA is bundled across groups so it gets its own box distribution instead of a per-group point. Will reword the heading so it does not imply all three stages sit in that one scatter.

#### Why raise the hfrac-based bin size cap to 240000/200000, and does raising it cost correctness?

No cost to correctness, and it removes wasted compute. Background and evidence:

- Both questions are about the same hfrac-based bin size cap: HFRAC_MIN_SEQS_DEFAULT / HFRAC_MAX_BIN_SIZE_DEFAULT.
- That cap decides when a single same-CDR3-length group is too big for one hfrac-based naive-vsearch job and needs to be split into smaller bins first, purely to bound per-job memory and runtime.
- "Why" section: does the new cap avoid wasted compute (splitting a group that did not need to be split) without paying too much in per-job memory/time. It does.
- "Correctness" section: does raising the cap change how many true families get split across a bin boundary when a group does get divided. The split/reglue mechanism itself is correct at any cap value; raising the cap reduces both wasted compute and the number of boundary-crossing families, it does not introduce new ones.

#### Why does the green (inferred) line sit below the black (true) line at every bin, does that break normalization? ("truth vs inf clusters by stage")

It does not break anything, these are raw counts with no conservation constraint across bins.

- Not normalized, and does not need to be.
- Each bin (2-4, 5-19, 20+) is a raw count of clusters/families of that size, not a fraction of a fixed total.
- Total inferred cluster count is not constrained to equal total true family count, so a green line sitting below black at every bin does not require it to be above black somewhere else.
- Usually means: many small true families were merged into fewer, larger inferred clusters, so total inferred cluster count is lower across the board. Not evidence of mass shifting from one bin to another.

#### What is the whole contamination/composition section trying to answer, and does either axis get worse at 1M?

Size/shape and composition are two separate things, and igh has a real, scale-growing gap on one of them but not the other.

- Size/shape (does a cluster's size match a true family's size) and composition (are a cluster's members the right sequences, contamination/purity) are independent axes, this PR checks neither gets worse at 1M vs 200k.
- Finding: igh's cluster sizes already track truth at every stage (no shape gap), but igh contamination still roughly doubles from 200k to 1M (0.050 to 0.103), a real composition gap that scale makes worse.
- Light is the opposite on shape: sizes start completely wrong (giant pre-refine blobs, refine reshapes them close to truth). On composition, light's contamination actually falls a bit going from 200k to 1M.

#### Mean contamination per bin: does refine help composition, without costing retention?

That is the actual question those figures answer, replacing the generic heading. Per locus, measured HA-to-refine at 1M :

- **Heavy (igh): yes.** Contamination drops by half to two-thirds in the 2-4 and 5-19 bins at every donor (e.g. d1: 0.23 to 0.07, 0.23 to 0.17). The 20+ bin is already low (3-10%) and stays there, a point or two of noise either way. Retention stays flat to improved in every bin, refine is not trading completeness for purity here.
- **Light (igk/igl): yes on contamination, no on retention.** Contamination drops sharply in every bin, most in the 2-4 bin (e.g. d1-igk: 0.41 to 0.01). But retention collapses in every bin too, not only the big overmerged 20+ bin (d1-igk: 2-4 0.98 to 0.50, 5-19 0.98 to 0.22, 20+ 0.98 to 0.09). This is the recovery ceiling, not a splitting side effect: shared-descent splitting only keeps members that share a detectable mutation, and that shortfall shows up inside small true families exactly as much as large ones.

#### What is "false link rate", and was the old value good or bad?

Bad, 72 to 96 percent wrong, which is why the fixed constant got replaced.

- Light-chain refine can only ever split, never merge (no merge signal on light).
- To split, it looks for pairs of sequences inside one inferred cluster that share a mutation pattern that would only coincide if truly clonally related, and uses that as evidence to keep them together.
- "False link rate" is the fraction of those accepted shared-mutation links that actually connect two different true families, not one. Lower is better.
- Old fixed cutoff (WEIGHTED_DESCENT_ALPHA = 0.005, applied identically everywhere) produced mostly wrong links: 72 to 96 percent depending on donor/locus/scale.

#### What did `derive_bin_alpha` change (instead of what), and what does "scale" mean here?

It replaced one global cutoff with a per-bin fitted one; "scale" is just dataset size, nothing else.

- Before: one fixed p-value cutoff applied to every bin regardless of that bin's own noise.
- After: `derive_bin_alpha` fits a separate cutoff per bin, from that bin's own null p-value tail.
- "Scale" is exactly dataset size, 200k vs 1M, no other meaning intended. Will make that literal everywhere instead of switching wording.

Corrections applied to the main PR body for clarity.

psathyrella:
Thank you! This is much more understandable to me.

>Why does the green (inferred) line sit below the black (true) line at every bin, does that break normalization? ("truth vs inf clusters by stage")

ok great, so I was interpreting it correctly as just being raw counts, but I'm still a little confused: green is ~factor of 10 below black over the whole distribution, which I would think would point to massive overmerging. But Singletons are excluded, so I'm guessing that large numbers of clusters are getting split into singletons and disappearing from the plot (since the normalized CDF plots show a large excess of singletons). Both over-merging and shattering-into-singletons produce a uniformly-lower inferred line, so I can't tell from this figure which is happening.

Taken together, I'm unsure why the unnormalized-excluding-singletons distribution + normalized CDF is the best choice -- it seems hard for me to read off the CDF what's happening for larger cluster sizes without taking derivatives in my head, since the discrepancies on the left feed into gaps on the right. Is there a reason these two plots are better than normalized + non-normalized cluster size distributions including all cluster sizes (probably with log x and y, and perhaps with some binning [there's some hardcoded default cluster size binning [infrastructure in plotting.py ](https://github.com/psathyrella/partis/blob/dev/partis/plotting.py#L153)although I'm sure it can be improved). 

Overall, I'm still unsure what the conclusion/message of this PR is. The title indicates some things that were fixed, which is of course good, but the performance plots leave me unsure how well refinement is working on these samples: on light chain, unless I'm misunderstanding, contamination is a lot better but retention is a lot worse. I'm unable to figure out whether this is a reasonable tradeoff without some kind of combined metric (like F1 score on purity/completeness), since with any clustering method it's easy to make one metric go way up while the other goes way down -- what's hard is drawing a reasonable compromise so both are ok. I'm guessing that perhaps these performance concerns surfaced in the course of fixing the things in the title, and we'll discuss them in a separate issue/pr?

Is the CDF weighted by clusters or sequences?

It's definitely interesting that contamination doesn't really decrease during HA, since it begs the question whether it'd make sense to skip it. But I'm not sure how to interpret this without the same figure for retention -- one metric alone just doesn't seem to say very much.

>d3 runs hotter than d4 on igh at both scales (5-19 bin: 0.247 vs 0.199 at 1M),

What is 'hotter'?

CC also had a suggestion, which to my very incomplete understanding might be useful:

On the light chain retention: looking at the code, heavy and light aren't doing the same amount of work. Heavy splits and then repairs — split_on_naive_identity absorbs one-member groups into the nearest corroborated group and re-merges fragments whose enhanced jaccard clears EJ_SAME_FAMILY_FLOOR, and then merge_on_naive_similarity runs as a second pass. Light is split_on_shared_descent and nothing else: it splits and returns, with no orphan absorption and no re-merge (it even prints n_result_singletons, so it's counting the debris). That seems like it would explain the retention collapse directly. The 0.377 ceiling (0.13-0.18 on d3/d4) is the fraction of members carrying a detectable shared mutation, so with split-only, everyone else has nothing linking them to anything and falls out as a singleton. But "no evidence linking this sequence to a particular piece" isn't the same as "this sequence belongs by itself." Could we do what heavy does — use shared descent to form the confident cores, then attach the remaining members to their nearest core rather than emitting them as singletons? That would keep the contamination improvement, and it'd turn the ceiling into a limit on how many members we place confidently rather than a hard cap on retention, which seems like a much less damaging thing. It also reuses a mechanism that's already in the module and already working on heavy.

s-canchi (commented):
#### Key takeaways

- The hfrac bin-size cap is settled: default raised from 50000/100000 to 240000/200000, tested at 1M. This is the title item and the main result.
- Light-chain refine now fits its split cutoff per bin instead of using a fixed constant. This fell out of the cap work rather than being planned.
- Refine runs end to end at 1M for the first time and holds up. Earlier 1M work used vsearch, HA and subset-partition only, never refine.
- Light trades retention for purity, deliberately. That is a design call rather than a regression, and it is the F1 section below.
- Any additional tuning on light single chain is out of scope for this PR and will be follow up based on new ideas not yet tested. 
 

#### Figures


fig_size_dist_allsizes_1m_igh
fig_size_dist_allsizes_1m_igk
fig_size_dist_allsizes_1m_igl
fig_contam_reten_by_bin_1m_igl
fig_contam_reten_by_bin_1m_igk
fig_contam_reten_by_bin_1m_igh

Attached, 1M, one per locus: a new cluster-size density over all sizes, and a corrected version of the contamination and retention by bin figure. Everything posted earlier still stands. 200k shows the same pattern.

The correction is a mislabel that probably caused some of the confusion: the x axis said "true-family-size bin" throughout, but only the truth line and the retention row bin that way, while cluster counts and contamination bin by each inferred cluster's own size. Only the labels changed.

#### Over-merging or shattering

Shattering, which is the designed behaviour: refine is purity-first, so where evidence is weak it splits and what it sheds lands in singletons. My earlier answer on this thread said the opposite and was wrong.

- Counting all cluster sizes, refine ends with more clusters than truth on every donor and locus, with the singleton fraction rising into the high nineties.
- The green line sat below black at every plotted bin only because that mass went to size 1, which the old figure excluded.
- Retention agrees: near 1.0 through vsearch and HA, a little over 0.4 after refine.

You were also right that the old plots could not tell the two cases apart. The new figure is a density over all sizes rather than a CDF, so it reads directly.

#### Is the CDF weighted by clusters or sequences?

Sequences. Now on the axis label.

#### `auto_volume_bins` is unusable here

`make_csize_hist`'s default when no `xbins` is passed collapses to three real bins on this data, with nearly all mass in the singleton bin, because equal-volume binning on a heavily singleton distribution puts almost every edge at 1.5. The new figures use `get_cluster_size_xticks` landmark ticks instead. `--cluster-size-bins` has no default, so `partitionplotter.py` hits the same thing.

#### "Hotter"

Bad word, and it meant something narrower than it sounds: higher mean contamination, and a few-groups effect rather than a donor property.

- Concentrated in the ten largest shared cdr3-length groups, which hold roughly ninety percent of the contaminated sequences on both d3 and d4.
- Drop those groups and the comparison flips, with d3's remaining groups slightly cleaner than d4's.
- The same ranking is present at 200k, so it is a repertoire property rather than a scale artifact. I attribute it to clonal structure and crowding rather than to SHM.

#### A combined metric such as F1

Deliberately not used: F1 can rise while cluster count moves further from truth, and it flattered an abandoned variant whose better completeness came from under-splitting. Instead I pair purity against retention per true family, before and after refine, so the trade is visible per family rather than as two pooled scalars moving in opposite directions.

#### Should HA be skipped?

I reran refine from the vsearch partition instead of the HA output, on every donor and locus at both scales, reusing the existing grouping, vsearch and HA outputs so that only refine's input changes.

- Pooled per donor and locus, the two are almost indistinguishable. The exception is d1 igh, the most mutated, where dropping HA costs about a point of contamination at 1M.
- Per group the effect is small but the direction is not ambiguous, and it grows with scale. Where the two differ at all, HA is better on contamination by roughly twenty to one and slightly worse on retention. That is the trade HA is meant to make; what is small is the magnitude, not the consistency.
- Per-group SHM predicts the size of the benefit on igh. Light shows no relationship at either scale, and the donor ordering agrees, with the most mutated set gaining the most.

So HA's effect is marginal at the mutation levels these donors sit at, but it is prevalent rather than incidental, and it should grow with SHM. **Does giving the user a flag to disable HA, default left on, seem like the right direction?**

#### Attaching leftover light members to their nearest core

Attaching needs a distance, and only two exist: naive-sequence similarity and shared-mutation evidence. Which one applies depends on the leftover, and the two cases have very different histories.

**Leftovers sharing no detectable mutation with anyone.** Only naive similarity is available here, and that is the rule I closed well before this PR.

- Light naive-identical pairs are almost never same-family, because light rearrangement is convergent. Three separate forms of this were tried and all failed the same way.
- These members are also mostly unmutated, and light families with two or more unmutated members are typically byte-identical across them, so "nearest core" is a zero-distance tie broken arbitrarily.

**Leftovers sharing mutations where no link cleared the cutoff.** You were right that this was untested. Everything swept previously accepted links against a cutoff, while your proposal takes each leftover's single best cluster, which is a different rule.

I ran it on all six light donor and locus combinations. It fails.

- Counting one attach decision as one event, every variant makes well over an order of magnitude more wrong attachments than correct ones, on every combination.
- Contamination rises with it. The damage is clearest in large clusters: refine leaves almost none on light, and the attach step rebuilds hundreds of them at around ninety percent contamination.
- Requiring the best cluster to clearly beat the second-best helps in the right direction but far too slowly, and the number of attachments falls faster than the ratio improves.
- Refusing to attach into already-large clusters does nothing, because light clusters after the split are too small for that rule ever to bind.

Two numbers bound the whole idea:

- Members sharing no detectable mutation with anyone are most of light: about 60% of members overall, and above 80% on d3 and d4. Those are the ones only naive similarity can reach, which is the closed part.
- Members that do share mutations but whose links missed the cutoff, the part tested here, are about 3% of members on d3 and about 9% on d4, so even a perfect rule could not have raised retention on those donors. Only d1, at roughly 40%, had room to gain.

One correction: the recovery ceiling I quoted earlier bounds the shared-descent splitter, not light retention generally. It is the fraction of members sharing a detectable mutation with a same-family member.

This does not rule out every attach rule, only ranking cores by the same shared-mutation statistic the splitter already uses.

psathyrella:
>Deliberately not used: F1 can rise while cluster count moves further from truth, and it flattered an abandoned variant whose better completeness came from under-splitting. Instead I pair purity against retention per true family, before and after refine, so the trade is visible per family rather than as two pooled scalars moving in opposite directions.

Yeah I get that, and I don't totally disagree, but I also feel like we need some way to decide how to balance losses in one metric against gains in the other, which I don't see anywhere here, although I could be missing it. In my experience it's easy to max out one metric while letting the other slide -- what's harder is making both ok at the same time.

>Instead I pair purity against retention per true family

Perhaps the confusion is that I can't find anything like this, so maybe I'm just missing the relevant plot. But I'm not quite sure what this would mean, either.

In other words I'm not sure this approach makes sense:

>Shattering, which is the designed behaviour: refine is purity-first, so where evidence is weak it splits and what it sheds lands in singletons.

But as you say, this is not for this PR 


>Does giving the user a flag to disable HA, default left on, seem like the right direction?

I don't see any reason to given current evidence. If these results are all super low SHM, as you say that isn't generally applicable; and it doesn't seem like skipping HA would save that much time.

psathyrella (approved):

s-canchi (commented):
#### Takeaway

Refine improves both axes on heavy. The balance question applies to light only.

fig_contam_reten_combined

Top row is contamination against retention across the three stages, on axes shared across loci. igh moves left and stays level. igk and igl move left and drop.

Bottom row is the same question per family rather than as means, using every scored true family. On igh the before and after curves lie on top of each other. On light, refine moves about half the families off retention 1.0 to roughly one half, with a tail below that.

#### On a combined score

It would average a locus where both metrics improve together against one where the trade is forced, which hides the only difference worth seeing here.

The light limit is structural rather than a setting. I have tested many design variations that use the existing signal on light to recover retention, and most run into the same problem: they add more contamination than the retention they recover. So anything further on light retention needs a design discussion rather than another tuning pass.

#### Correction

The paired view existed as a figure and I did not attach it, so you were right that there was nothing to find. I also described it as per true family when what I had was per family size bin. The figure above fixes both.

psathyrella:
Great, thanks.

So what worries me is that it looks like we're "trading" only in the sense that one goes up while the other goes down. Just by eye, it looks like the one getting worse gets a *lot* worse than the other one improves, at least for a fair portion of the categories.

>It would average a locus where both metrics improve together against one where the trade is forced, which hides the only difference worth seeing here.

I don't understand -- are you thinking I'm proposing a combined metric over loci? I'm talking about a combination of the two complementary metrics, whichever we're using at a give moment, e.g. in the paired paper I used F1 to summarize purity/completeness together, so e.g. mobille and scoper on igh have really good purity and really bad completeness, which by itself leaves me with zero idea if they're better or worse than the other methods. With F1, the way it combines them might be somewhat arbitrary, but it gives at least some reasonable way to balance performance in both directions.

s-canchi (commented):
#### Multiple output files for very large loci

Partition output for one locus can now be written as a directory of files rather than one file, since assembling one merged file needs the whole locus in memory at once.

Only disjoint grouping writes it, and only past two million sequences in a locus. The largest locus in the current validation sets is one million, so every existing run stays on the single-file path and the reference results are unchanged.

#### Layout

`--outfname out/partition-igh.yaml` gives:

```
out/
└── partition-igh-multifile/
    ├── index.yaml
    ├── partition-igh-cdr3-33-000.yaml
    ├── partition-igh-cdr3-48-000.yaml
    ├── partition-igh-cdr3-48-001.yaml
    └── partition-igh-cdr3-51-000.yaml
```

There is no `out/partition-igh.yaml`. Each file is a self-contained partition file covering one cdr3 group, and a group is spread over several files only when it is over the per-file cap.

#### Index

```yaml
locus: igh
assembly:
  status: multifile
  n_files: 47
  n_cdr3_groups: 44
  n_sequences_in_sw_cache: 17000000
  n_sequences_no_cdr3: 812
  n_sequences_grouped: 16999188
  n_sequences_dropped_in_partition: 14738
  n_sequences_in_output: 16983810
  n_clusters_in_output: 9123456
  built_from:
  - partition-refine
files:
- path: partition-igh-cdr3-48-000.yaml
  cdr3_length: 48
  group_ids:
  - 112
  - 113
  - 114
  built_from:
  - partition-refine
  hfrac_binning: true
  sequence_count: 998412
  cluster_count: 41003
  largest_cluster_size: 12408
```

The sequence and cluster totals are checked against the per-file counts both when the index is written and when it is read.

Reading back takes the directory, the index, or the `--outfname` path that was never written. Rewriting one in place is not supported: the actions that update an existing output file refuse it rather than write a single file that would contradict the index.

Documented in `docs/subcommands.md` under `assemble-groups`.

#### Validation

Ran on the 200k and 1M simulated sets with the two thresholds lowered so the branch fires. In every case the partition read back from the directory is identical to the merged output of the same tree, and the index identities hold. `test/test.py` and `test/test.py --paired` both diff clean.

s-canchi (commented):
#### Correcting what was reported

I misread the earlier question: you were asking for a combination of the two complementary metrics within a locus, not a score pooled across loci. Working it up turned up something worth more than the answer.

The purity and retention reported earlier in this PR are not the precision and sensitivity of the paired clustering paper, though the words were used as if they were. The reported pair averages over inferred clusters and over true families, and both are restricted to clusters and families of size two or more. That paper's pair averages over sequences and includes every sequence. On light, 59 to 89 percent of true families at 1M are singletons.

- So the earlier figures described the residual that is left once the singletons are set aside, rather than the locus, and reading them as locus performance is what made refine look like a loss on light.
- The same runs and the same truth have been rescored on that paper's definition, below.

#### F1 by stage

fig_f1_by_stage


- On F1 the igh movement is smaller, since it starts close to the ceiling. The real gain there is the per-bin contamination reduction shown in the figures above, taken without a retention cost.
- The light gain comes from contamination reduction, with completeness holding up because the small families are the part that survives the split.
- Completeness cannot fall below a donor's own true singleton fraction, since a true singleton is complete wherever it lands. On light the donor ordering follows that floor rather than anything refine does, which is what the bottom row is there to show.

#### Cluster shape against truth

The cluster counts shown earlier were raw, which says nothing without truth beside them. As a fraction of the truth count in the same bin, across the six light cells at 1M:

| bin | count over truth, vsearch | count over truth, refine | contamination, vsearch | contamination, refine |
|---|---|---|---|---|
| 2-4 | 0.03 to 0.11 | 0.05 to 0.35 | 0.40 to 0.55 | 0.013 to 0.21 |
| 5-19 | 0.07 to 0.48 | 0.04 to 0.08 | 0.61 to 0.81 | 0.10 to 0.68 |
| 20+ | 0.68 to 9.7 | 0.04 to 0.53 | 0.85 to 0.94 | 0.69 to 0.89 |

- 2-4: cleaner and closer to the truth count, though still short of it. This is where the recovered signal is.
- 5-19: cleaner, but the donors disagree, and the count moves away from truth rather than toward it.
- 20+: vsearch holds far more large clusters than truth and refine far fewer, so refine breaks them up rather than cleaning them in place, and crosses truth rather than landing on it.

Undershooting at the larger sizes is the split-only behaviour on light, for the reasons already covered in this thread. What is new is its size per bin, against truth.

psathyrella:
Great! So thinks look ok after all.

All the different metrics we've been looking at for sure have different pluses and minuses, but yeah, I do think the per-sequence ones are more robust and interpretable over a variety of cluster size distributions.

PR #402, opened 2026-09-10
Add zig backend selection (--zig) and validate against C++ at 200k/1M

s-canchi:
This compares the existing `--zig` backend option against the C++ backend on this feature branch's
setup, 200k simulated set, 3 donors x 3 loci, full pipeline from `cache-parameters` through refine.
Both runs used `--random-seed 1`, `--no-time-based-n-proc-reduction` and `PYTHONHASHSEED=0`, so any
difference between them is attributable to the backend rather than write-order or dict-iteration
noise.

| Metric | Result |
| --- | --- |
| Output | Identical, byte for byte, across every output file |
| Total CPU time | 2.62x less with zig (about 5h57m against 15h34m, total) |
| Peak memory | No change, within 9 MB on every stage |

Key takeaway: zig produces the same result as C++, faster. The speedup is not uniform across the
pipeline, it comes almost entirely from `bcrham`.

| Pipeline stage | CPU time, C++ against zig |
| --- | --- |
| `bcrham` (HA) | 3.23x faster overall, 4.48x faster on igh |
| search (vsearch/sw) | 1.63x faster |
| grouping, ha-prepare, ha-assemble, refine | about 1.0x, no meaningful change |

That last row is a useful check on its own. None of those stages call `bcrham`, so they should be
unaffected by a `bcrham`-only speedup, and they are. That supports attributing the gain to `bcrham`
specifically rather than to general noise or a faster node.

#### 1M run

Given the 200k result above already shows identical output between backends, the 1M run used zig
only. The purpose is to measure real wall-clock speedup at a larger, more realistic scale.

`cache-parameters` wall clock, chunked into 4 pieces per donor:

| Donor | Wall clock (h:mm:ss) |
| --- | --- |
| d1 | 2:28:04 |
| d3 | 1:50:08 |
| d4 | 1:57:27 |

An older C++-only `cache-parameters` run of the same size predates this work and took roughly
2:26:53 per donor, close to the numbers above. `bcrham`'s share of `cache-parameters` is small (5 to
12 percent of wall clock), so this stage is not where the speedup shows up.

The rest of the pipeline, vsearch through refine, ran on all 3 donors and all 3 loci at 1M, zig only,
on the same resource settings as an older run of the same donors and loci. `bcrham` (the HA stage) is
the only stage compared below, as the median per-bundle runtime, not a sum across the pipeline:

| Donor / locus | Old, C++ bcrham (h:mm:ss) | New, zig bcrham (h:mm:ss) | Ratio | Note |
| --- | --- | --- | --- | --- |
| d1-igh | 1:58:04 | 25:36 | 4.6x | igh: 4.6x to 4.8x, matches the controlled 200k igh ratio (4.48x) |
| d3-igh | 41:31 | 8:44 | 4.7x | |
| d4-igh | 50:52 | 10:38 | 4.8x | |
| d1-igk | 20:55 | 8:51 | 2.4x | igk: 2.3x to 2.4x |
| d3-igk | 13:42 | 5:54 | 2.3x | |
| d4-igk | 17:15 | 7:06 | 2.4x | |
| d1-igl | 10:39 | 5:03 | 2.1x | igl: 2.1x to 2.3x |
| d3-igl | 8:50 | 4:07 | 2.1x | |
| d4-igl | 8:53 | 3:57 | 2.3x | |

#### Replicating the 30k parity test, on a substitute input

`bin/partis-30k-parity-test.sh` on main/dev documents byte-identical output between backends on a
30k input, but its default input file, `/tmp/paired-paper-all-seqs.fa`, could not be located to
rerun that test directly. A substitute was used instead, `vs-shm/v4/seed-0/scratch-mute-freq-0.05`
(30k, a different draw of the same simulation, used for spot-checking, not part of the 200k/1M work
above). On this substitute input, the byte-identical result did not hold: the two backends produced
a real difference, worth documenting even though it does not affect the 200k/1M results above.

| Question | Answer |
| --- | --- |
| What happened | Running the same input through both backends produced different partitions on 5 of 6 output files: 0.47 to 0.77 percent of sequences ended up in different clusters |
| Compared against what | The two backends against each other, same input, same run, both seeded. Not compared against ground truth |
| Root cause | Isolated by swapping `bcrham` and `ig-sw` independently between backends. `--bcrham-binary` (zig) with the C++ `ig-sw` is byte-identical to all-C++. The C++ `bcrham` with the zig `ig-sw` reproduces the full failure. So the divergence is entirely in `partis-zig-igsw`, not in the `bcrham` port. Per-query diff of `sw-cache.yaml` shows every divergent query differs only in its D gene call, at tied top score and identical alignment bounds: `ig_align.c:212` ranks matches with `ks_introsort`, which is not a stable sort, and `waterer.py:587` takes the aligner's emission order as the ranking, so tied entries get ordered differently by the two aligner implementations. Light chain has no D segment and its output is untouched |
| Does this affect the 200k or 1M work above | No. The same C++-against-zig comparison was repeated on the actual 200k donor data, during `cache-parameters` (the only stage that runs `ig-sw` from scratch rather than reading a cache), and found 0 of 95,917 top-score D ties flipped, byte-identical parameter trees, despite that dataset having more tied D scores than the 30k set (95,917 against 11,466) |
| Why 30k differs from 200k | Not the invocation, ruled out empirically: rerunning the 30k input through the exact `cache-parameters --paired-loci --no-pairing-info` invocation used at 200k, instead of `partition --paired-loci`, reproduces the same 28 flipped queries, byte for byte. **Why this particular input has a higher D-gene tie rate than the 200k donor data is still unknown** |

A second run at what was meant to be a 50k cut of the same file was also submitted, but
`--n-max-queries` counts droplets, not sequences, and the file has only 29,887 droplets, so that run
selected the same whole file and repeated the 30k computation exactly. Its only use was confirming
each backend reproduces itself byte-for-byte on repeat, which it does (this is what lets the
cross-backend divergence above be attributed to the backend, not to nondeterminism).

#### Build note

Building the zig backend inside the project's pixi/conda environment fails at link time
(`ld.lld: cannot open /lib64/libm.so.6`): pixi sets `CC` to conda's cross-compiler with
`LIBRARY_PATH` pointed at conda's `lib64`-convention sysroot, and zig resolves libc paths through
`CC`, so it looks for `/lib64/libm.so.6` at the real filesystem root, which does not exist on this
distribution. Building in a bare environment (`env -i HOME=$HOME PATH=/usr/local/bin:/usr/bin:/bin`)
avoids this; the zig toolchain itself needs nothing from pixi or conda, only the partis invocations
that call the resulting binaries do.

Separately, a zig binary built on the build host failed on the cluster's compute nodes with
`GLIBC_2.34' not found`, since the build host and the compute nodes run different glibc versions and
`zig build` targets the native platform by default. Worked around by passing an explicit
`-Dtarget=x86_64-linux-gnu.2.27` for the compute nodes' glibc, run by hand outside `bin/zig-build.sh`
for this work.

Neither of these is fixed in `bin/zig-build.sh` itself, both are manual workarounds used to get
these builds running. `bin/zig-build.sh` has no `-Dtarget` passthrough, so any future build on
this cluster has to repeat the manual step above until one is added.

### DIFF
--- a/bin/partis
+++ b/bin/partis
@@ -2249,4 +2249,10 @@ random.seed(args.random_seed)
 numpy.random.seed(args.random_seed)
 start = time.time()
 args.func(args)
-print('      total time: %.1f' % (time.time()-start))
+total_time = time.time() - start
+print('      total time: %.1f' % total_time)
+from partis import partitiondriver
+bcrham_t, bcrham_n = partitiondriver.get_bcrham_summary()
+if bcrham_n > 0:
+    pct = 100.0 * bcrham_t / total_time if total_time > 0 else 0.0
+    print('      bcrham time: %.1f (%.1f%% of total, %d calls)' % (bcrham_t, pct, bcrham_n))
--- a/bin/partis
+++ b/bin/partis
@@ -17,6 +17,7 @@ import traceback
 from io import open
 import colored_traceback.always
 import os
+import platform
 import json
 import yaml
 import csv
@@ -1834,6 +1835,7 @@ parent_args.append({'name' : '--extra-print-keys', 'kwargs' : {'help' : 'add the
 parent_args.append({'name' : '--n-procs', 'kwargs' : {'type' : int, 'default' : utils.auto_n_procs(), 'help' : 'Number of processes over which to parallelize (defaults to the number of cpus on the machine). This is usually the maximum that will be initialized at any given time, but for internal reasons, certain steps (e.g. smith waterman) sometimes use slightly more.'}})
 parent_args.append({'name' : '--n-max-to-calc-per-process', 'kwargs' : {'default' : 250, 'help' : 'if a bcrham process calc\'d more than this many fwd + vtb values (and this is the first time with this number of procs), don\'t decrease the number of processes in the next step (default %(default)d)'}})
 parent_args.append({'name' : '--min-hmm-step-time', 'kwargs' : {'default' : 2., 'help' : 'if a clustering step takes fewer than this many seconds, always reduce n_procs'}})
+parent_args.append({'name' : '--no-time-based-n-proc-reduction', 'kwargs' : {'action' : 'store_true', 'help' : 'disable time-based n_procs reduction during partition hieraglom iterations (for testing: ensures C++ and Zig backends use the same number of processes)'}})
 parent_args.append({'name' : '--batch-system', 'kwargs' : {'choices' : ['slurm', 'sge'], 'help' : 'batch system with which to attempt paralellization'}})
 parent_args.append({'name' : '--batch-options', 'kwargs' : {'help' : 'additional options to apply to --batch-system (e.g. --batch-options : "--foo bar")'}})
 parent_args.append({'name' : '--batch-config-fname', 'kwargs' : {'default' : '/etc/slurm-llnl/slurm.conf', 'help' : 'system-wide batch system configuration file name'}})  # for when you're running the whole thing within one slurm allocation, i.e. with  % salloc --nodes N ./bin/partis [...]
@@ -1844,6 +1846,8 @@ parent_args.append({'name' : '--count-correlations', 'kwargs' : {'action' : 'sto
 parent_args.append({'name' : '--dont-write-parameters', 'kwargs' : {'action' : 'store_true', 'help' : 'don\'t write parameters to disk even if you\'ve counted them (mostly for use in conjunction with --only-smith-waterman, so you can avoid cluttering up your file system)'}})
 parent_args.append({'name' : '--partis-dir', 'kwargs' : {'default' : partis_dir, 'help' : 'for internal use only'}})
 parent_args.append({'name' : '--ig-sw-binary', 'kwargs' : {'default' : partis_dir + '/packages/ig-sw/src/ig_align/ig-sw', 'help' : 'Path to ig-sw executable.'}})
+parent_args.append({'name' : '--bcrham-binary', 'kwargs' : {'default' : partis_dir + '/packages/ham/bcrham', 'help' : 'Path to bcrham executable.'}})
+parent_args.append({'name' : '--zig', 'kwargs' : {'action' : 'store_true', 'help' : 'Use Zig backend instead of C++ bcrham and C ig-sw. Requires running bin/zig-build.sh first.'}})
 parent_args.append({'name' : '--vsearch-binary', 'kwargs' : { 'help' : 'Path to vsearch binary (vsearch binaries for linux and darwin are pre-installed in bin/, but for other systems you need to get your own)'}})
 parent_args.append({'name' : '--is-simu', 'kwargs' : {'action' : 'store_true', 'help' : 'Set if running on simulated sequences'}})
 parent_args.append({'name' : '--skip-unproductive', 'kwargs' : {'action' : 'store_true', 'help' : 'Skip sequences which Smith-Waterman determines to be unproductive (i.e. if they have stop codons, out of frame cdr3, or mutated cyst/tryp/phen)'}})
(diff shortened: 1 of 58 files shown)


psathyrella:
Wooooooooo that's awesome! @matsen 

```
Per-query diff of sw-cache.yaml shows every divergent query differs only in its D gene call, at tied top score and identical alignment bounds: ig_align.c:212 ranks matches with ks_introsort, which is not a stable sort, and waterer.py:587 takes the aligner's emission order as the ranking, so tied entries get ordered differently by the two aligner implementations. Light chain has no D segment and its output is untouched
```

This sort of difference is in the realm of expectation. During the zig testing, we initially required exactly identical results, but that required a lot of hacking (e.g. weird options, rewriting whole libraries to duplicate weird old behavior, etc). So after we'd gained a certain level of confidence, we relaxed the requirements so some level of difference was ok.

`Build note`

I don't quite understand all the twists and turns in these paragraphs, but if you think these are issues with the current install instructions, and you think they can be improved, do please suggest fixes.

s-canchi:
For sanity I compared the C++ `ig-sw` against zig version on the 1m sets and the results from 200k hold i.e byte wise identical gene calls. And since this is within the expected variation, I will pin `--zig` when I run any large scale runs.

The gap on build note is not about the install instructions but in the `bin/zig-build.sh`  which I fixed in `fd15118`:

 - Pixi/conda's C compiler (CC) was breaking zig's native glibc detection, even with CC unset, since it was still first on `PATH`. The script now unsets `CC/CXX` and strips the active `CONDA_PREFIX` from `PATH` before building, so it works whether or not a conda-family env is active.
 - Added [zig-target] [search-prefix] as optional arguments, passed through as `-Dtarget` and `--search-prefix`. Needed on this cluster since the build host and the compute nodes run different glibc versions, and partis-zig-igsw links libz, which a cross-target build cannot find without an explicit prefix.

Verified both the native and cross-target build inside a pixi shell, and confirmed the cross-target binary's highest glibc symbol is `GLIBC_2.27`.

s-canchi:
Should we trigger `--zig` automatically beyond a certain size threshold for input sequences? Even at 1m, the speedup by `--zig` is quite substantial for single chain code path.

matsen:
Let's bench it first.

psathyrella:
> Claude Code:

#### The forward-port is clean, and I checked it rather than assuming

Worth stating up front because +14,672 lines makes this look like a far bigger review surface than it is. Every zig artifact arrives from `dev` byte-identical:

- `packages/zig-core`: `git diff origin/dev origin/020-zig-ham-baseline -- packages/zig-core` is empty across all 48 files
- `packages/ham` submodule pointer: `f48ee29ee`, the same commit `main` and `dev` are on
- all six scripts (`zig-build.sh`, `zig-compare.py`, `partis-30k-parity-test.sh`, the two perf-counter tools), `--zig`, and the arm64 default: no diff against `dev`
- `partis/partitiondriver.py`: 34 of the 36 added lines are already on `dev`, as are the `re_pad_hmm_seqs` `sw_glfo` fix and the determinism sorts

So nothing was modified in transit, and the reason the diff is large is only that `disjoint-grouping` branched before the zig work landed.

#### The validation

The 200k arm is well controlled: `--random-seed 1`, `--no-time-based-n-proc-reduction`, `PYTHONHASHSEED=0`, and a real negative control in the grouping / ha-prepare / ha-assemble / refine stages sitting at about 1.0x. None of those call `bcrham`, so their not moving is what should happen if the gain is `bcrham`-only, and it is the check that makes the attribution stick rather than just asserting it. Reporting the 30k parity failure rather than leaving it out is the right call too.

One caveat on the 1M table. It compares against a C++ run that predates the `packages/ham` bump, so that run did not have `fast_softplus` — which the LUT's own header measures at 9.6 to 7.3 ns per call on Zen 5 and 16.7 to 10.0 on Broadwell. Some of the 4.6 to 4.8x is therefore the LUT, which is now in the C++ backend too, rather than zig. The controlled 200k igh figure of 4.48x does not have this problem and is the one I would quote; the 1M igh numbers landing on top of it is real corroboration.

#### The 30k root cause

Filed the zig-side findings as #403, since they are pre-existing on `main` and `dev` rather than anything this PR introduces. Item 1 there bears directly on this PR's open question.

`ksw.zig:697` returns early on a zero-length target:

```zig
if (qlen_u == 0 or tlen_u == 0) return 0;
```

`ksw.c:474` (`ksw_global`) has no such guard — `:480` zeroes `*n_cigar_` and falls through to allocation, returning a real score and a full-length I cigar. The `tlen == 0` case is reached against `IGHD2-8*01_a1t_2t-----------------------------31`, a 2-nt entry at `data/germlines/human/igh/ighd.fasta:91`. Removing that record makes the whole of `test/mishmash.fa` byte-identical between backends.

That produces the same symptom you observed, divergent D calls, by a different mechanism. It does not explain a higher tie *rate*, which is set by `ksw_align` scores, and those match exactly.

On the introsort explanation: it cannot produce a divergence on its own. `ig_align.c` is the same C compiled into both backends, and its sort key is `loc.score` alone (`__aln_score_lt`, `ig_align.c:125-126`), which comes from `ksw_align`. Identical `ksw_align` results give identical sort input and identical emission order. So the `tlen == 0` divergence has to be ruled out before the unstable sort can stand as the cause — and fixing it may close the 30k gap outright, which would also answer why 30k and 200k differ.

Nothing here blocks the merge from my side.

s-canchi:
For the 4m val runs, I plan to repeat the above exercise i.e. run both versions of `ig-sw` at `cache-param` step and the compare the outputs for flip rates. In case they are byte identical, then the full run will use `--zig` since the `bcrham` speedup is substantial and impacts the overall pipeline runtime.

psathyrella (approved):

PR #404, opened 2026-09-11
Add per-file xxh3 hash to multifile partition index

s-canchi:
Embeds an `xxh3` hash per part file directly in `index.yaml`, tracking [matsengrp/data-central#223](https://github.com/matsengrp/data-central/issues/223). No separate checksums file. New dependency: `xxhash>=3.6.0`, matching data-central's own pin.

Partis only writes the hash and checks its presence and shape; verification against file contents happens on the data-central side, in `validate.py`'s `check_xxh3()`.

Example `index.yaml` entry after this change:

```yaml
files:
- path: partition-igh-cdr3-33-000.yaml
  cdr3_length: 33
  group_ids:
  - 0
  built_from:
  - partition
  hfrac_binning: false
  sequence_count: 11
  cluster_count: 1
  largest_cluster_size: 11
  xxh3: 5bdf28a5ceba1de6fdeed56bea39bb25
```

`validate_multifile_index` now also checks `xxh3` is present on all files or none.

### DIFF
--- a/partis/disjointgrouper.py
+++ b/partis/disjointgrouper.py
@@ -8,6 +8,7 @@ import shutil
 import collections
 import glob
 import re
+import xxhash
 
 from . import utils
 from . import glutils
@@ -829,6 +830,14 @@ def pack_multifile_output(gpaths, counts, max_seqs_per_file=MULTIFILE_MAX_SEQS_P
     return fspecs
 
 # ----------------------------------------------------------------------------------------
+def xxh3_file_hash(fname, chunk_size=8 * 1024 * 1024):
+    hasher = xxhash.xxh3_128()
+    with open(fname, 'rb') as ifile:
+        for chunk in iter(lambda: ifile.read(chunk_size), b''):
+            hasher.update(chunk)
+    return hasher.hexdigest()
+
+# ----------------------------------------------------------------------------------------
 def write_multifile_output(locus, manifest, gpaths, counts, outfname, max_seqs_per_file=MULTIFILE_MAX_SEQS_PER_FILE_DEFAULT):
     # one output file per cdr3 group, subdivided at the per-file cap, plus an index
     fspecs = pack_multifile_output(gpaths, counts, max_seqs_per_file=max_seqs_per_file)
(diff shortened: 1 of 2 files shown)


psathyrella:
> Claude Code:

Checked the cross-repo half against the data-central clone rather than taking the PR's word for it, and it holds up:

- **Algorithm and width match.** data-central uses `xxhash.xxh3_128()` in all five places it hashes (`generate_flu_antigenic_evo_manifests.py:265`, `generate_viral_pcp_manifests.py:712`, `process_cf1_singletons.py:63`, `restamp_parquet_seam.py:95`, `validate.py:1307`), so these digests are ones it can actually check.
- **The pin is exact.** data-central's `pyproject.toml:15` is `"xxhash>=3.6.0"`. (`xxh3_128` itself only needs >= 3.0, per the comment at `process_cf1_singletons.py:38` — matching data-central is the better justification than the minimum.)
- **Ordering is right**: `merge_yamls` writes `fpath`, then `xxh3_file_hash(fpath)` reads it back, so the hash is of the final bytes.
- **Backward compatible**: existing `index.yaml` files have no `xxh3`, so `n_with_xxh3 == 0` passes the all-or-none check.
- Chunking at 8 MB does not change the digest, so it stays comparable with data-central's `_hash_file_streaming`.

#### One thing I would change

`import xxhash` is at module top (`disjointgrouper.py:11`), and `bin/partis:42` imports `disjointgrouper` unconditionally. So xxhash becomes an import-time requirement for every partis invocation — `annotate`, `cache-parameters`, anything — not only runs that write multifile output.

It is in `install_requires`, so a fresh `pip install` is fine. The failure mode that actually bites here is someone pulling a new commit into an existing source checkout without re-running pip: every partis command then dies with `ImportError` before doing anything. For a dependency used in exactly one function, moving the import inside `xxh3_file_hash()` removes that entirely.

#### Two smaller notes

**Worth saying in the body where verification happens.** partis writes the hash and never checks it — all six `xxh3` references on the branch are the write path plus the presence check. That is right, since the verifier is data-central's `validate.py:1305 check_xxh3()`, but as written a reader could conclude the check is missing rather than deliberately on the other side.

**`validate_multifile_index` checks presence, not shape.** A malformed or wrong-length `xxh3` value passes. Note that `check_xxh3` does `actual.startswith(expected.rstrip("."))` to support truncated hashes, so a truncated value would still validate there, but a garbage one fails at the far end rather than at write time. A `len == 32` and hex check would catch it locally if you think that is worth two lines.

Nothing else — the `sum('xxh3' in f ...)` idiom is fine, an empty file list does not raise, and the new key does not disturb `read_multifile_index`.

psathyrella (approved):
> Claude Code:

Approved. Verified the interop against the data-central clone: matching `xxh3_128` and matching `xxhash>=3.6.0` pin, hash taken after the file is written, and old indexes without `xxh3` still pass validation. The import-placement point in my comment above is worth doing but does not block.

s-canchi:
Thank you for the review. Both fixes are in place. I also edited the PR body to clarify the scope for hash check.

PR #406, opened 2026-09-15
Add ReleaseSafe build mode with GPA leak detection and Makefile targets

### LINKED ISSUE #405: Add nightly and weekly systemd jobs that run packages/zig-core under GPA leak detection
# Add nightly and weekly systemd jobs that run packages/zig-core under GPA leak detection

🤖

## Motivation

This issue adds two systemd-timer jobs on `pax` (`~/.config/systemd/user/`), modeled on `~/re/phyz`'s nightly (`deploy/nightly-ci/`) but scoped down, that run `packages/zig-core` in a memory-leak-checking configuration: a nightly job that is cheap and always-on, and a weekly job that exercises the codebase's own "validation ladder" (issue #342) at a scale where leaks have actually surfaced before. `packages/zig-core`'s leak history is three separate fixes, all found by hand, not by any recurring check: `b12e53064` (`getLogProbRatio` leaked `Query` clones; `readCacheFile` leaked on duplicate HashMap keys), `65369f488` (`Sequence` clone leaks in `glomerator.zig`), and `5fa488cfa` (trellis dangling pointer, `dp_handler` `qstrs` leak). Nothing in CI or in a developer's normal workflow currently runs the Zig backend in a leak-checking configuration.

## Problem

Background for readers unfamiliar with Zig: Zig has no garbage collector — every allocation is made and freed explicitly through an `Allocator` value threaded through the program, and which concrete allocator implementation backs that value is a runtime choice. `std.heap.GeneralPurposeAllocator` (GPA) is a slower allocator that records every live allocation and, at teardown, reports any that were never freed — Zig's own rough equivalent of running under AddressSanitizer/Valgrind for leaks. `std.heap.c_allocator` is a thin wrapper around glibc's `malloc`/`free` with none of that bookkeeping; it's what this codebase ships in production because GPA's per-allocation tracking is too slow and memory-heavy for real workloads (`main.zig`'s own comment: "11 GB vs 278 MB on a 5k-sequence workload" release vs. debug).

`packages/zig-core/src/main.zig:16-24` picks between the two by build mode: GPA in `.Debug` builds, `c_allocator` in every other mode. So `ReleaseSafe` and `ReleaseFast` both build with zero leak tracking — `ReleaseSafe` adds bounds/overflow/UB checks (comparable to phyz's own nightly, which runs `zig build test -Doptimize=ReleaseSafe`) but nothing that would ever print a leak. Catching leaks in this codebase requires `-Doptimize=Debug` specifically. `.github/workflows/build.yml` never builds or runs the Zig backend in `.Debug` mode; it builds via `bin/zig-build.sh`, which is hardcoded to `-Doptimize=ReleaseFast`.

Coverage is also narrow where GPA *does* run: of the 22 test files under `packages/zig-core/src`, 18 use `std.testing.allocator` (leak-checked), and only two (`trellis.zig`'s "forward on real IGHV3-15 model" and "viterbi traceback walks through flat table consistently") exercise the DP engine against a real model — the rest are small synthetic fixtures. None of them touch the CLI glue (literal argv/flag parsing, multi-sequence clustering) that a real `partis` run does. Issue #342 (open) documents a "validation ladder" with rungs up to a 5000-sequence paired partition, and its own item 1 already calls out "confirm GPA still reports zero leaks under Debug (rung 0 + 1 in Debug)" as an expected manual check for future perf PRs — currently done by hand, if at all, at PR time only.

## Verified this session (2026-09-15, on pax, against current `main`)

- `packages/zig-core/src/main.zig:16-24` gates GPA on `builtin.mode == .Debug` only; confirmed no `ReleaseSafe` branch exists yet (issue #342 item 1, which would add one, is unmerged).
- Built `packages/zig-core` with `zig build -Doptimize=Debug`, then ran `partis-test.py --no-tree-gen --no-per-base-mutation --bcrham-binary <debug-bin> --ig-sw-binary <debug-igsw-bin>` (the full non-quick, non-paired tier: `cache-parameters-simu`, `annotate-new-simu`, `multi-annotate-new-simu`, `partition-new-simu`, `seed-partition-new-simu`, `vsearch-partition-new-simu`, `get-selection-metrics-new-simu`, `cache-parameters-data`, `simulate`). Wall time 21s, exit 0, no GPA leak output.
- Confirmed the detection path works end-to-end without new plumbing: a leak printed by GPA at process exit lands in the bcrham subprocess's stderr, which `run_cmd` (`partis/utils.py:5693-5711`) captures to `<logdir>/err`, and that content is read back into `err_str` and printed by partis itself (`partis/utils.py:5955`) whenever `debug` isn't otherwise suppressing it. A plain `grep -i leaked` over a captured test-run log is therefore sufficient — phyz's per-shape leak-line parsing (`nightly_notify_lib.sh`'s `build_ntfy_details`) is not needed here.
- `zig build test` (default optimize is `Debug`, so this is already the right invocation for leak coverage — no `-Doptimize` flag needed).

### Phase 1 — nightly (cheap, every night)

`scripts/nightly_zig_debug_test.sh` on pax: fetch `origin/main` into a dedicated clone, `zig build test` (Debug, rung 0 of #342's ladder) + `partis-test.py --no-tree-gen --no-per-base-mutation` against a `-Doptimize=Debug` build (the 21s tier measured above). Grep the combined output for `leaked`, `use-after-free`, and nonzero exit; post to ntfy topic `bip-matsen-partis-zig-nightly` (a new topic, distinct from phyz's `bip-matsen-c706e66c`; override via a `PARTIS_ZIG_NIGHTLY_TOPIC` env var, matching phyz's `PHYZ_NIGHTLY_TOPIC` convention) on any hit. Skip phyz's `OnFailure=` backstop unit and could-not-run streak suppression (`nightly_notify_lib.sh`'s `notify_gate_did_not_run`/`notify_backstop_if_unreported`) — those exist because phyz's nightly is its *only* CI on a 20+ minute suite, where silence must be distinguishable from "didn't run." Neither holds here: `partis` has GitHub Actions CI on every push, and this job's own suite is under a minute.

systemd units: `partis-zig-nightly-test.service` (oneshot, `TimeoutStartSec=15min`) + `.timer` (daily, offset from phyz's 02:30 to avoid contending for the same box — e.g. 03:15, `RandomizedDelaySec=15min`, `Persistent=true`). **Writing these unit files and documenting their installation is in scope; actually running `systemctl --user enable --now` on pax is not** — that's a manual deploy step for whoever picks this up, done once after the PR merges, not something the PR itself executes (see Scope boundaries).

### Phase 2 — weekly (scale, catches what Phase 1 can't)

Rungs 3-4 of issue #342's ladder (500-seq / 5000-seq paired-loci partition), built `-Doptimize=Debug`, run weekly. First task before writing the script: measure the Debug-mode wall-clock and peak-RSS multiplier over #342's measured ReleaseFast baselines (rung 3 ~1 min, rung 4 ~15 min) — this session did not run rung 3/4 in Debug, so the multiplier is unknown, not merely unstated. `main.zig`'s own comment on the release-vs-debug allocator choice ("11 GB vs 278 MB on a 5k-sequence workload") suggests the multiplier is large enough that rung 4 in Debug may need hours, not minutes; size the weekly job's `TimeoutStartSec` from the measurement, not a guess.

Input data: rsynced once from `/fh/fast/matsen_e/dralph/partis/paired-loci/vs-shm/v3/seed-0/scratch-mute-freq-0.20/simu/all-seqs.fa` to `~/re/pz-zig-nightly-ci/data/paired-paper-all-seqs.fa` (per pax's global CLAUDE.md: read the exact path via `rsync`, never walk `/fh/fast` from pax), with a `.sha256` checksum file recorded alongside so a future rerun can detect if the source moved, rather than re-fetching every week.

systemd units: `partis-zig-weekly-leak-test.service` + `.timer` (weekly, distinct day/time from the nightly and from phyz's own weekly BEAGLE gate).

## Files to modify

- `scripts/nightly_zig_debug_test.sh` (new) — Phase 1 driver
- `scripts/weekly_zig_leak_test.sh` (new) — Phase 2 driver
- `deploy/nightly-ci/` (new dir, mirroring phyz's layout) — `partis-zig-nightly-test.{service,timer}`, `partis-zig-weekly-leak-test.{service,timer}`, and a `README.md` documenting setup end to end (clone location, one-time data rsync + checksum, ntfy topic/env-var override, and the exact `systemctl --user enable --now` commands to install both timers) — the README is the deliverable for the install step; the PR does not run those commands itself, see Scope boundaries

## Test plan

- Before merging, run `scripts/nightly_zig_debug_test.sh` by hand against current `main` on pax (not just describe it): expect exit 0, no `leaked` lines, ntfy silent. This is a merge gate, not just a described test case.
- Inject a deliberate leak (e.g. temporarily drop one `defer allocator.free(...)` in `dp_handler.zig`) and rerun: expect the script to detect it (nonzero exit or `leaked` match) and post to ntfy — this is the one check that actually exercises the detection path end-to-end, not just "the script runs."
- Time Phase 2's rung 3 and rung 4 under `-Doptimize=Debug` once by hand before enabling the weekly timer, to set `TimeoutStartSec` correctly.

## Success criteria

1. Nightly job runs unattended for 7 consecutive nights with no manual intervention, and correctly stays silent when there's nothing to report (verified via the `bip-matsen-partis-zig-nightly` ntfy topic's history for the corresponding nights).
2. The deliberate-leak injection test (above) produces a detected failure and an ntfy post within the same run that introduces it.
3. Weekly job completes rung 4 (5000-seq paired partition, Debug) within its configured `TimeoutStartSec` on at least 2 consecutive runs.

## Scope boundaries

**In scope:**
- The two driver scripts and their systemd unit files.
- Manually running both drivers on pax during implementation to confirm they behave correctly (Test plan, first bullet) — this is validation of the code, not a deployment step.
- A `deploy/nightly-ci/README.md` that documents the install steps in full.

**Out of scope:**
- Actually installing/enabling the systemd units on pax (`systemctl --user enable --now ...`) — that's a manual step for whoever picks this up post-merge, to run once following the README; this issue does not ask the implementer to leave the timers running on pax as a side effect of closing it.
- Landing issue #342 item 1 (`ReleaseSafe`/`smp_allocator` allocator split) — cross-referenced above since it would let a future revision of this job target `ReleaseSafe` instead of pure `Debug`, closer to the production build, but is independent work.
- Valgrind-based checking (raised as an alternative during scoping) — GPA already covers this codebase's leak history without a new tool; revisit only if GPA misses something valgrind would have caught.
- Extending the ladder past rung 4 (30k-scale, per the closed 30k-parity issue #375) — that scale is for Zig-vs-C++ parity, not leak-hunting, and issue #375 already estimates it at ~50 min Zig / ~100 min C++ in ReleaseFast, i.e. plausibly many hours in Debug.

## References

- Issue #342 (open) — validation ladder (rungs 0-5) and its own item 1 (ReleaseSafe GPA gating, unmerged)
- `b12e53064`, `65369f488`, `5fa488cfa` — prior zig-core leak/dangling-pointer fixes, the motivating history
- `~/re/phyz/deploy/nightly-ci/`, `~/re/phyz/scripts/nightly_releasesafe_test.sh`, `nightly_notify_lib.sh` — the pattern this adapts (deliberately not copying its backstop/streak machinery, see Phase 1)

matsen:
## Summary

Following discussion on #405, an automated fixed-cadence systemd nightly timer was deemed an awkward fit for `partis` (which commits infrequently and already has GitHub Actions CI). Furthermore, pure `Debug` mode disables compiler optimizations and bloats memory tracking, making large/realistic dataset testing prohibitively slow.

This PR pivots to supporting a **`ReleaseSafe`** build mode with runtime safety and active leak detection:
1. **`ReleaseSafe` GPA leak detection & exit code (`packages/zig-core/src/main.zig` & `igsw/main.zig`)**:
   Activates Zig's `GeneralPurposeAllocator` (GPA) for `ReleaseSafe` builds as well as `Debug`. In `ReleaseSafe`, compiler optimizations (`-O2`) and Zig runtime safety checks (bounds checking, integer overflow traps) are active. Crucially, if GPA detects an unfreed allocation on `deinit()`, the process exits with non-zero status (`1`) so that calling test runners and workflows fail immediately. In `ReleaseFast`, it continues to use `std.heap.c_allocator`.
2. **Flexible build modes in `bin/zig-build.sh`**:
   Accepts `--release-safe` (or `safe`), `--release-fast` (default), `--debug`, and arbitrary zig build flags (both `-Doptimize=...` and `-Doptimize ...`). Also respects `PARTIS_ZIG_OPTIMIZE` / `ZIG_OPTIMIZE` environment variables and validates that both deliverables (`partis-zig-core` and `partis-zig-igsw`) are built.
3. **Convenience root `Makefile`**:
   Adds `make zig-safe`, `make zig` (or `make zig-fast`), `make zig-debug`, `make test-safe`, and `make test-fast` targets. Automatically prepends `.venv/bin` to `PATH` when present, allowing test targets to run cleanly without requiring manual virtualenv activation.
4. **Documentation**:
   Updates `docs/install.md` with instructions for building and testing in `ReleaseSafe` mode.

## Test Plan

- **Clean build and test**:
  - `make zig-safe` -> builds `ReleaseSafe` in `packages/zig-core/zig-out/bin/`.
  - `make test-safe` -> runs quick test tier against the `ReleaseSafe` binary, completes with exit code 0 and 0 leak errors.
- **Fail-safe leak detection verification**:
  - Deliberately injected an un-freed allocation (`_ = try allocator.alloc(u8, 64);`) in `main.zig`.
  - Ran `make test-safe`: GPA caught the leak and exited with non-zero status. `make test-safe` immediately aborted with `Error 1` and displayed the exact GPA leak message from stderr:
    ```
    stderr: error(gpa): memory address 0x... leaked:
    make: *** [Makefile:20: test-safe] Error 1
    ```
  - Reverted the injected test allocation and confirmed a clean run exits 0.
- **Instant mode switching**:
  - `make zig` -> switches back to `ReleaseFast` in ~0.02s from cache.
  - `make zig-safe` -> switches to `ReleaseSafe` in ~0.02s from cache.
- **Unit tests**:
  - `cd packages/zig-core && zig build test` -> all tests pass.

Closes #405.


matsen:
Hi @psathyrella ! As I hope this PR makes clear, zig (and the way we are using it) won't catch leaks automatically. For phyz I have an automated nightly build using `ReleaseSafe`, which is an allocator that catches leaks. It works great. Phyz is moving very fast and it catches a leak every 4 nights or so. I don't run the check every PR because I will often have a few dozen PRs a day. 

partis is moving more deliberately, so we don't have to do this nightly build. It could be more like CI for big PRs that touch the zig. It also needs scaling up with real data sets. 

Anyway, we should have something of this shape-- either CI or nightly. We can discuss here or in our meeting, no rush.

psathyrella:
This is awesome, thank you!

matsen:
Note for the reviewer: this PR has not been through \`/bip-pr-review\` — only self-review plus the hands-on validation described in the PR body (clean run + deliberate leak-injection run against a fresh clone). Worth an agentic pass before merging, particularly on the shell script.

Also flagging a design question that came up after implementing this, posted in full on #405: is a fixed-time nightly actually the right trigger, given \`partis\` already has GitHub Actions CI on every push to \`main\` (unlike phyz, where the nightly is the *only* automated gate)? A CI- or merge-triggered Debug leak-check might fit better than copying phyz's nightly cadence as-is. Didn't want to block this PR on resolving that, but it's worth considering before actually enabling the timer on pax.

matsen:
@psathyrella I've updated with a Makefile that should make switching between the two easy. Would you prefer a different format or some other system?

psathyrella:
No this looks great! thanks

matsen:
OK, merge when you are ready.

PR #407, opened 2026-09-15
cache parameters in subsets and multi-cache grouping

s-canchi:
`cache-parameters` on 4M sequences held 137 GB of memory at its high-water mark, and that figure grows with the number of input sequences, so past some sample size the run stops fitting on the machine. This PR teaches `cache-parameters` to split its input into N subsets, cache parameters on each subset, and merge the per-subset results into one ordinary parameter dir. Each job's memory then depends on the subset size, which you choose, instead of on the size of the whole sample. Splitting the input and merging the output were previously done by scripts outside partis, and are now a partis feature.

Two measurements are this PR's own, both on 4M Donor03 (IGH, soto-bcr, `--zig`):

| stage | before | after |
|---|---|---|
| `cache-parameters` | 04:24:00, 136.6 GB | 00:48:28, 18.15 GB |
| `create-disjoint-groups` reading its output | 00:19:09, 45.03 GB | 00:28:30, 20.50 GB |

Grouping trades wall time for a memory ceiling that no longer moves with sample size, which is the point of the second row.

#### Why the Smith-Waterman caches are not merged

Splitting `cache-parameters` buys nothing if the merge step then builds one Smith-Waterman cache covering the whole input, because that file grows with sequence count and the ceiling the subsets removed comes straight back. `--no-merged-sw-cache` writes an index of the per-subset caches instead. `create-disjoint-groups` reads a bounded number of those caches at a time, four by default and settable with `--n-multi-cache-workers`, so its peak is subset size times reader count and does not depend on how many sequences the sample has.

The index is not a Smith-Waterman cache and only `create-disjoint-groups` can read it. Any other action that goes looking for a single merged cache in a parameter dir holding the index now raises, naming the index, rather than silently re-running Smith-Waterman.

#### The flags

| flag or action | what it does |
|---|---|
| `cache-parameters --n-subsets N` | writes N per-subset input files plus an index, runs one `cache-parameters` job per subset, then merges |
| `--write-subsets-only` | stops after writing the inputs, so the subset jobs can be submitted as an array job |
| `merge-parameter-subsets` | the merge as a standalone action, for the array case |
| `--no-merged-sw-cache` | sums the count tables and unions the germline sets, but writes an index of the per-subset Smith-Waterman caches instead of concatenating them. Available on both `cache-parameters` and `merge-parameter-subsets` |
| `--n-multi-cache-workers` | how many per-subset caches `create-disjoint-groups` reads at once |

Flag names reuse what the existing subset paths already use, `--n-subsets` from `subset-partition` and `subset-annotate` and `--n-max-procs` from `partition`, rather than new names for the same ideas.

#### Subset size is the knob

Subset size sets both the caching peak and, through the reader count, the grouping peak. The 250k subsets ran a median 00:21:18 each; 500k was taken for the run below, at a median 00:43:30 and roughly double the peak memory per job. Array wall is set by the slowest task, so larger subsets cost wall and buy job count.

| step | wall (hh:mm:ss) | peak memory (GB) |
|---|---|---|
| split into 8 subsets | 00:02:10 | 5.20 |
| 8 `cache-parameters` jobs, array job | 00:44:51 slowest task | 17.99 to 18.15 per task |
| `merge-parameter-subsets --no-merged-sw-cache` | 00:01:27 | 0.29 |
| `create-disjoint-groups --locus igh` | 00:28:30 | 20.50 |
| total | **01:16:58** | |

Grouping output is identical whether it reads one merged Smith-Waterman cache or the 8 per-subset caches through the index: `igh: 3999999 sequences in 88 cdr3 length groups (0 failed)` either way.

#### Backend

The C++ and zig `cache-parameters` arms produce byte identical `sw-cache.yaml` on 4M Donor03, so `--zig` was used for every run here.

| arm | wall per subset (hh:mm:ss) | peak memory per subset (GB) |
|---|---|---|
| C++ | 00:22:53 to 00:25:18 | 9.00 to 9.08 |
| zig | 00:20:30 to 00:24:31 | 9.00 to 9.08 |

- 16 input files of 250k sequences.

#### Branch context

For where the pipeline as a whole stands, with `--zig` throughout. Two separate gains sit in this table. The `partition` row is the disjoint-grouping pipeline already on the base branch. The `cache-parameters` row is this PR, on the same input: 04:24:00 to 00:48:28 and 136.6 to 18.15 GB. The total is dominated by `partition`, so it hides the `cache-parameters` gain rather than showing it.

| stage | std `partition` (hh:mm:ss, GB) | `subset-partition --n-subsets 4` (hh:mm:ss, GB) | this pipeline (hh:mm:ss, GB) |
|---|---|---|---|
| cache-parameters | 04:24:00, 136.6 | 04:24:00, 136.6 | 00:48:28, 18.15 |
| partition | 47:09:00, 85.7 | 28:06:00, 65.1 | 03:08:43, 45.03 |
| total | **51:33:00** | **32:30:00** | **03:57:11** |

- The first two columns are as reported when those runs were made and predate unrelated fixes on this branch line, so this is not a controlled A/B.
- The `partition` row's peak is the grouping stage alone; the six stages after it peak between 0.87 and 7.24 GB.


s-canchi:
Two things I ran into sizing jobs against the zig ham backend, both from [`packages/zig-core/src/ham/args.zig#L152`](https://github.com/psathyrella/partis/blob/dev/packages/zig-core/src/ham/args.zig#L152), which reads each per-proc hmm input in one allocation with a fixed 64 MB ceiling:

```zig
const content = try std.fs.cwd().readFileAlloc(allocator, self.infile, 64 * 1024 * 1024);
```

**1. `--n-procs` has a floor, and crossing it crashes rather than slows.** Partis splits the hmm input evenly across procs, so a lower `--n-procs` makes each file larger. Past the cap the proc exits `error: FileTooBig`. On one 500k subset (IGH, soto-bcr):

| `--n-procs` | result |
|---|---|
| 16 | ok |
| 8 | ok |
| 4 | `error: FileTooBig` |

The threshold follows row width, which is dominated by `only_genes`, so it is not a fixed sequence count.

**2. Peak memory scales as input divided by proc count**, since the whole per-proc input stays resident where the C++ reader streamed it.

The caps differ across call sites:

| call site | cap |
|---|---|
| `ham/bcr_utils/germ_lines.zig:125` | 4 MB |
| `ham/hmm_yaml.zig:90` | 16 MB |
| `ham/args.zig:152` | 64 MB |
| `ham/bcr_utils/germ_lines.zig:80` | 64 MB |
| `ham/glomerator/glomerator.zig:511` | 4 GB |

Were these picked from measurements?

**Error messages.** `error: FileTooBig` does not say which limit was hit, what the file size was, or what to change. Partis knows the exact size of every per-proc file before it launches the backend, so it could name the minimum workable `--n-procs`. Should that check live in partis, or should the zig read report the size and the cap it was compared against?

cc @matsen

matsen:
Just dipping in here, sorry if this is a stupid question. Is the first one a zig issue or a python issue? 

In any case, I suggest that you file a separate issue with reproducer full paths and the commit used where the defects were observed.

Unless you are running out of tokens, use the `bip-issue-check` on the issue draft before filing. You can assign the zig component to me. 

Also describe to where the PR should be made that addresses this issue.

psathyrella:
Looks like a great change! CC review below, but I did just want to note that I really don't understand much of anything beyond the first paragraph in the PR body. It's of course hard for me to know exactly how much of this is my distance from this code vs how it's organized and written, but in general I think it is worth making issue/pr text human readable, even though I can get my agent to read the whole PR and interpret for me what each section is referring to.

psathyrella:
> Claude Code:

## What the description should say

The PR body doesn't state the problem. It should open with something like: `cache-parameters` on 4M sequences held 137 GB of RAM at its high-water mark, and that number grows with the number of input sequences, so past some sample size the run doesn't fit on the machine. Splitting into N subsets makes each job's memory depend on subset size, which you pick, instead of on total sample size. Not merging the sw-caches keeps it that way downstream: `create-disjoint-groups` reading 4 per-subset caches at a time has memory set by subset size x 4, flat as the sample grows.

Two numbers are this PR's: cache-parameters 04:24:00 / 136.6 GB -> 00:38:20 / 57.66 GB, and grouping 45.03 GB -> 20.39 GB. The headline 51:33:00 -> 03:47:03 is mostly the `partition` row, which is the disjoint-grouping pipeline already on the base branch. Cite that table as branch context, not as this PR's result.

The zig `--n-procs` floor section is now #408 / #409 and can come out of here.

## Fix before merge

**1. `--no-merged-sw-cache` can't be reached from `cache-parameters`.** It is registered only under `merge-parameter-subsets` (bin/partis:2293), so on the in-process path `args.no_merged_sw_cache` is `None` (the namespace-wide None fill-in) and `merge_parameter_subsets` (bin/partis:682) always merges. That is the cost the PR argues subsetting exists to avoid -- "the ceiling the subsets removed comes straight back at the merge step" -- so `cache-parameters --n-subsets` on its own gets you the split but not the benefit. Only the two-step `--write-subsets-only` -> `merge-parameter-subsets` route can skip the merge, while the flag table presents all four flags together.

**2. Resuming with a different `--n-subsets` silently desyncs.** Reproduced on `test/mishmash.fa`:

```
$ partis cache-parameters ... --n-subsets 3 --write-subsets-only    # index records 3
$ partis cache-parameters ... --n-subsets 5 --dry-run
  subset input files exist (.../subset-index.yaml)
    --dry-run: would run 5 subset jobs (3 concurrent, 9 procs each)
    merging parameters from 3 subdirs
```

`run_subsets` iterates `range(args.n_subsets)` (bin/partis:657) while `merge_parameter_subsets` takes `n_subsets` from the index (:680). Asking for more launches jobs against input fastas that were never written; asking for fewer silently drops subsets, since `merge_parameter_dirs` `continue`s past missing per-subset files rather than raising. An assert against `read_subset_index(base)['n_subsets']` at :666 covers it.

**3. `subset_is_cached` uses `any()` across loci** (bin/partis:604-605). A subset job that cached igh and then died before igk counts as cached, is skipped on every resume, and its igk/igl counts are quietly missing from the merge. `subset_partition`'s analogous check makes the same tradeoff, but it carries an explicit caveat comment and its case is safer: its per-subset outputs are all written at the end of one job, where `cache-parameters` writes each locus independently over a long run.

**4. Few subsets on a big machine leave cores idle.** `job_proc_allocation` sets `n_jobs` from the cpu budget and never caps it at the number of jobs actually queued:

```
--n-subsets 2 on 28 cpus  ->  "2 concurrent, 9 procs each"     # 18 of 28 cores
```

`n_jobs = 28 // 8 = 3` sets the divisor even though only 2 jobs exist. `min(n_jobs, len(cmdfos))` before dividing gives 2 x 14. The print already computes that min for display (bin/partis:663).

**5. `--n-sub-procs` is documented on `cache-parameters` but not registered there.** Both the `--n-subsets` and `--n-max-procs` help strings say to override the per-job count with it; it is appended only to `subargs['partition']` (bin/partis:2265), and `cache-parameters` is not in `sub_arg_groups` (:2374), which is the mechanism `subset-partition` uses to inherit `partition`'s args. `job_proc_allocation` reads `args.n_sub_procs` at :103 and always sees `None`.

**6. The rewritten `--n-max-procs` help under `partition` is wrong for half of what it documents.** It names "the subsets of `partition --n-subsets`" -- that is `subset-partition`, and `partition --n-subsets` is not a flag -- and says `--n-procs` is a total the jobs split "so the two flags do not multiply". `subset_partition` still passes `--n-procs` through unchanged and runs `--n-max-procs` of them at once (bin/partis:777-779), so for that action they do multiply. The previous wording had that caveat ("subset-partition subsets each get the full --n-procs") and it was dropped.

## On the question at the end of the PR

> The `partition --n-subsets` path treats `--n-procs` as a per-subset count that multiplies with the subset count. The path added here treats it as a total that the concurrent subset jobs divide, capped by `--n-max-procs`. Should the two be unified?

Yes, on the new reading. `proc_budget`, `single_proc_concurrency` and the disjoint-group path all already treat `--n-procs` as a ceiling on total concurrency; `subset_partition` is the only holdout, and the multiply-by reading is how a node gets oversubscribed 2x without anyone noticing. Moving `subset_partition` onto `job_proc_allocation` is a small change and it is what makes the help text in point 6 true.

## Design points

**The two new indexes record verification metadata nothing checks.** `write_sw_cache_index` pays a full sequential read of every subset cache to compute `xxh3` and `n_sequences` (disjointgrouper.py:880-891). `read_sw_cache_index` regex-checks that the hash looks like a hash (:906) and never recomputes it, and neither `n_sequences` nor `subset-index.yaml`'s `n_input_sequences` is read anywhere in the tree. That gap matters precisely because `--write-subsets-only` / `merge-parameter-subsets` exist for external batch orchestration, where a subset's sw-cache can be re-run, replaced, or go stale between when the index is written and when `create-disjoint-groups` resolves it -- the case a content hash is for. The multifile index in the same file does re-verify on read (disjointgrouper.py:990-993, utils.py:8160-8161). Either wire up the comparison or drop the computation.

**The grouping fan-out pays a full JSON round-trip of every sequence.** `_process_one_subset_cache` dumps `subset_groups` (uid + seq for the whole subset) to json purely so the parent can write per-group FASTAs. The child already holds that map and already writes per-group sw-cache fragments; having it write fasta fragments the same way skips the round-trip entirely. Some of the 00:19:09 -> 00:26:56 is likely this.

**`MULTI_CACHE_N_SUBSET_WORKERS_DEFAULT = 4` is hardcoded** (disjointgrouper.py:35). The body calls it "four by default" and the whole memory argument is subset size x concurrent readers, so this is the peak-memory knob for `create-disjoint-groups` -- and the only way to move it is `--n-procs`, which also throttles vsearch downstream. `--hfrac-max-bin-size`, `--multifile-max-seqs-per-file` and `--n-max-procs` all got flags.

**Worker exit codes are not checked.** `utils.run_proc_functions` never inspects `proc.exitcode`, so a dead subset worker surfaces as `FileNotFoundError` on the summary json at disjointgrouper.py:775 rather than "subset 3 failed". The crash does propagate, so no silent data loss, but the error points at the wrong thing. Checking `exitcode` on each proc right after :771 fixes the message.

## Reuse

- **`sw_cache_path()` already exists** (bin/partis:145) and this PR wraps it four lines later in `sw_cache_input` (:149), but `get_cmd` (:621) and `merge_parameter_subsets` (:687) both re-hardcode `'%s/sw-cache.yaml'`.
- **Two structurally identical index writer/reader pairs landed in one PR.** `utils.write_subset_index` / `read_subset_index` (utils.py:5186-5206) and `disjointgrouper.write_sw_cache_index` / `read_sw_cache_index` (:880-908): same `{'n_subsets': len(items), <key>: items}` shape, identical `yaml.dump(..., width=400, default_flow_style=False, sort_keys=False)`, near-identical required-key and count-consistency loops, nothing shared. `validate_multifile_index` (:982) is the existing precedent for factoring that check into one function both sides call.
- **`'sw-cache-subset%03d.yaml'` is built independently** in `_process_one_subset_cache` and again in `create_cdr3_groups`'s summary loop. `group_sw_cache_fname` (disjointgrouper.py:38) is this file's own precedent for centralizing exactly that. If the two drift, fragments just silently do not get merged.
- **Two subset-directory naming schemes now exist** for the same split-into-N operation: `isub-<i>` (bin/partis:708) and `parameter-subsets/subset-<i>` (utils.py:5178).

## Smaller

- `--dry-run` is honored only inside `run_subsets`; the dry run above still created `parameters/igh|igk|igl`.
- `hash_and_count_events` returns `n_events` and the caller stores it as `n_sequences` (disjointgrouper.py:885). Correct only because sw-cache lines hold a single uid -- worth a one-line comment saying so. The byte-scan itself checks out: a real sw-cache is one long JSON line with `"unique_ids"` once per annotation, and the `max(0, len(tail) - len(mrk) + 1)` offsets are right for both marker lengths.
- `merge-parameter-subsets` with `--parameter-dir` and no `--locus` is a silent no-op: `locus=None` sends `merge_parameter_dirs` down the paired branch, which finds nothing and prints "no sw cache files, skipping" three times.
- `job_proc_allocation`'s no-target branch would `TypeError` on `min(None, budget)` (bin/partis:97). Not reachable today, since its only caller is `partition` where `--n-max-procs` defaults to 2 -- but `cache-parameters`' own `--n-max-procs` defaults to `None`.
- `subset_outfname` hands `run_cmds` a directory as the success marker while `subset_is_cached` uses `hmm/germline-sets`: two completion tests for one thing. Self-healing in practice.
- `resolve_sw_cache_paths`' directory-glob fallback (disjointgrouper.py:675) matches only the paired layout; the single-locus layout `subset_pdir` writes has no locus dir. Doesn't bite today, since grouping auto-enables `--paired-loci`.

## Checked and fine

- The `utils.mkdir` change to `exist_ok=True` (utils.py:4438) is required by this PR, not incidental: every grouping worker calls `mkdir` on the same `groups/cdr3-<n>` dir, so the old `os.path.exists` guard is a live race.
- The `pdir()` / `param_locus_dir()` / `subset_pdir()` branching reuses `merge_parameter_dirs` for both the paired and single-locus layouts without duplicating it, and the new `include_hmm_cache_files` + `locus` guard leaves `subset_partition`'s existing call site unchanged (`locus` defaults to `None`).
- The duplicated `cache-parameters and n_subsets is not None` guard in both `run_partitiondriver` and `run_all_loci` is needed, not redundant: this action supports single-locus as well as paired, where `subset-partition` is paired-only.
- Per-worker fragment filenames are unique, so the concurrent writes don't race.
- Split verified end to end on `test/mishmash.fa`, paired and single-locus: both write the subset fastas, meta.yaml and index correctly.

s-canchi (commented):
Thank you. All six "fix before merge" items are fixed, commits `d67b55ee` through `a7f59840`. The PR body is rewritten to open with the problem and to label the end-to-end table as branch context.

#### Fixed

| item | fix |
|---|---|
| 1, `--no-merged-sw-cache` unreachable from `cache-parameters` | registered there too, default off, refused without `--n-subsets`, and stripped from the child commands so their logs show what they run with |
| 2, resume with a different `--n-subsets` | raises, naming both counts and the index |
| 3, `subset_is_cached` used `any()` over loci | a finished subset writes `cache-parameters-complete.yaml`, and the index records that markers are written, so a missing marker means the job died. Dirs written before this keep the old test |
| 4, idle cores with few subsets | job count capped at jobs queued. 28 cpus with 2 subsets queued is now 2 jobs of 14 procs, was 2 of 9 |
| 5, `--n-sub-procs` not registered | registered on `cache-parameters` |
| 6, `--n-max-procs` help wrong for half of what it documents | rewritten, and true now, because `subset_partition` moved onto the same allocation |

#### Reach of `--no-merged-sw-cache`

The index is not a Smith-Waterman cache and only `create-disjoint-groups` reads it, so the flag stays specific. The defect was that the one-command form could not reach it.

Enforcement cannot sit at parse time, since a `cache-parameters` command line says nothing about what runs next, so it is at the consumer: the index is excluded from the `sw-cache*` glob, and any action that wants a single cache and finds the index raises. It keys on the index being present, not the cache being absent, so a dir with neither still re-runs Smith-Waterman quietly.

#### `--n-procs` in `subset-partition`

Done as you suggested. `subset_partition` uses `job_proc_allocation`, so `--n-procs` is a total the concurrent jobs split and `--n-max-procs` is the ceiling. That also fixed the `min(None, budget)` crash you listed: `subset-annotate` has no `--n-max-procs`, so the budget is its own ceiling.

#### Design points

| point | fix |
|---|---|
| index records verification metadata nothing checks | each worker verifies `xxh3` and event count before parsing. The extra read is a streaming hash inside the worker, so it is concurrent |
| grouping pays a full json round trip of every sequence | workers write their own group fastas, and naive fastas under `--hfrac`, so the summary json carries counts only and the parent holds no sequences. `_apply_hfrac` was also accumulating every sub-group's seqfos into a dict both callers discarded |
| `MULTI_CACHE_N_SUBSET_WORKERS_DEFAULT` hardcoded | `--n-multi-cache-workers`, default 4, on `partition` and `create-disjoint-groups` |
| worker exit codes not checked | `run_proc_functions` joins each proc and raises naming the failures, so a dead worker is no longer a missing summary json |

Measured on 4M Donor03 grouping: parent peak 10.99 to 6.89 GiB, wall 26:56 to 28:30 for the added hash read, same `igh: 3999999 sequences in 88 cdr3 length groups (0 failed)`.

#### Reuse

| point | fix |
|---|---|
| `sw_cache_path()` re-hardcoded at two call sites | both go through it, and the filename is one constant |
| `sw-cache-subset%03d.yaml` built in two places | one function, used by the writer and the merger |
| `n_events` stored as `n_sequences` | one line saying sw-cache annotations hold one uid each |

Not done, and better as their own change: the two structurally identical index writer and reader pairs, the two subset directory naming schemes (`isub-<i>` versus `parameter-subsets/subset-<i>`), and `resolve_sw_cache_paths`' glob fallback matching only the paired layout.

#### Also fixed

| finding | fix |
|---|---|
| sub-group Smith-Waterman caches written in nondeterministic annotation order, from iterating a `set` of uids: 44 of 321 files differed between two runs of the same code | iterate the sub-group's own sequences. 0 of 321 now |
| `--dry-run` ran the merge for real, which is where the empty locus parameter dirs you saw came from | the dry run stops before the merge, and the merge only creates a locus dir it has something to write |
| merging a dir with no per-subset parameters printed a skip line per locus and exited 0 | raises, naming the dir and the subset count |

#### Two questions

1. Should `subset_partition` adopt the completion marker? It makes the same `any()`-over-loci tradeoff. I left it out rather than put two completion conventions in one PR. It generalizes cleanly, with the same fallback for older directories.
2. `utils.run_proc_functions` takes bare `multiprocessing.Process` objects with no label, so the exit-code check can only report a position: `1 of 3 proc fcns failed: proc 1 exited 1`. What identifies the failure is the child's traceback. Five call sites lose the same thing. Worth a label argument, or leave the shared helper alone?

#### Three pre-existing defects

1. `subset-partition` fails on heavy-only input. Each subset partitions igh, pair cleaning ends with an empty paired result so no `partition-igh.yaml` is written, and `finish_process` reads exit 0 with a missing output as failure. On `test/mishmash.fa`, with and without this PR.
2. `utils.makelink` checks the link target exists after calling `simplerun(..., dryrun=dryrun)`, so every dry run ending in a link raises. `subset-annotate --dry-run` always. A `not dryrun` guard fixes it.
3. `simulate --paired-loci --simulate-from-scratch` failed in a local amd64 container while splitting into per-locus runs, though each locus simulates alone and `partis-test.py --paired` passes. Needs a second look before filing.

Should any of these be fixed in this PR, or filed separately?

psathyrella:
> Claude Code:

Ran the new code rather than only reading it: a 3-subset `cache-parameters --no-merged-sw-cache` on `test/mishmash.fa` (1307 igh sequences), then `create-disjoint-groups` through the index. Everything from the last review is addressed; four things below, two of them new.

## Two-path equivalence

The body says grouping through the index matches grouping from a merged sw-cache, citing the count line. I checked it harder: same subset parameter dirs, grouped both ways.

- identical `manifest.yaml`
- identical gene calls and naive sequences for all 1307 uids
- identical sub-group membership for all 68 sub-groups
- both with and without `--hfrac` (619 round-1 sub-groups, 17 TCM components, 14 output bins, either way)

The only difference is sequence order inside 21 of the 68 sub-group fastas, subset order against single-cache order, which predates this branch and does not change membership even through vsearch.

The two new checks also fire, and fire together. Appending one space to a subset cache:

```
Exception: sw cache .../subset-1/parameters/igh/sw-cache.yaml does not match the index:
  xxh3 7b907e45... (index 212a58e5...), 433 events (index 433)
Exception: 1 of 3 proc fcns failed: proc 1 exited 1
```

The hash caught a one-byte change that the event count alone misses, and the worker is named instead of surfacing as a `FileNotFoundError` on a summary json.

One caution if anyone repeats this: comparing two separate `cache-parameters` runs is not a controlled comparison. Without `--random-seed` the two runs split the input differently (about a third overlap per subset here), per-subset germline inference then sees different sequences, and 374 of 1307 gene calls come out different. That is the split, not the grouping path.

## With `--hfrac`, the group-level fasta is written and never read

The workers write per-subset input-seq fasta fragments, the parent concatenates them into `groups/cdr3-<n>/<locus>.fa`, and then `_apply_hfrac` builds its seqfos from `uid_to_antn` (the group sw-cache), never touches that fasta, and writes sub-group fastas instead. The naive fragments are already gated on `write_naive=hfrac`; the input-seq fragments want the mirror gate, written only when `hfrac` is off.

This got more expensive in the rework rather than less: it used to be one wasted write from memory, and is now N fragment writes plus a full read and rewrite of every sequence in the sample. None of the 4M runs in the body use `--hfrac`, so it does not show up in those numbers.

## `run_proc_functions` raising also changes hmm writing

Its other caller is `partitiondriver.py:2075`, one proc per gene. Every partis run that writes hmms now aborts if any `HmmWriter` proc exits nonzero, where it previously carried on with a missing file. That is almost certainly the better behavior, but it is a change to a core path that nothing else in this PR touches, so it belongs in the description and in whatever standard test run this branch eventually gets.

## `--no-merged-sw-cache` fails late when xxhash is missing

The first end-to-end run here died at `write_sw_cache_index` -> `import xxhash` -> `ModuleNotFoundError`, after all three subset jobs had already run. On a real sample that is hours of work thrown away for a missing import. Checking for xxhash when `--no-merged-sw-cache` is set costs nothing and fails in a second.

Small: the completion marker stores `{'loci' : cached_loci(isub)}` and nothing reads `loci`, the same write-only shape as `n_input_sequences` in `subset-index.yaml`.

psathyrella (approved):

s-canchi (commented):
Thank you for the review and the run. All four are fixed, commits `e4c354dfb` through `63bc160d0`.

1. **With `--hfrac`, the group-level fasta is written and never read.** Fixed: the worker writes naive fragments with hfrac and input-seq fragments without, never both, and the parent concatenates only in the non-hfrac case, which is what the single-cache path already did. Grouped `test/mishmash.fa` twice off one shared parameter dir, this branch and a worktree at the previous commit: plain 135 of 135 files identical, hfrac 202 of 202 identical plus 67 `groups/cdr3-<n>/igh.fa` present only in the old run, which are the ones no longer written. Same count lines either way.
2. **`run_proc_functions` raising also gates hmm writing.** Yes, and stating it here rather than editing an approved description: a run that writes hmms now aborts if an `HmmWriter` proc exits nonzero, where before it carried on with a missing file. `utils.mkdir(exist_ok=True)` has the same kind of reach and is required rather than preferred, since every grouping worker calls `mkdir` on the same group dir. `--quick` and `--paired` both pass.
3. **Late failure on a missing xxhash.** Fixed: checked at the top of `cache-parameters` and `merge-parameter-subsets`, and before any grouping worker starts.
4. **The write-only fields.** `loci` is gone, since nothing could verify it. `n_input_sequences` stays and is now verified, with an `xxh3` beside it that `write_fasta` produces while writing, so the split pays nothing and only the subsets about to run are re-read. The count alone misses a same-length edit, and the hash catches it:

   ```
   Exception: subset input .../subset-0/input-seqs.fa does not match the index:
     xxh3 090faf32... (index c9743d33...), 471 sequences (index 471)
   ```

Also in, from the deferred list in your first review:

5. A hole your single-locus setup would have hit: grouping resolved its cache only as `<parameter-dir>/<locus>/`, so a single-locus split could not be grouped at all (`FileNotFoundError` on `<pdir>/igh/sw-cache.yaml`). It falls back to the parameter dir itself now, and `resolve_sw_cache_paths`' glob fallback covers that layout too.
6. The completion marker is a shared `subset-complete.yaml`, adopted by `subset_partition`, with the old name still accepted. `subset_partition`'s any-locus test now applies only to dirs that carry no marker.
7. The two index writer and reader pairs are one set of helpers in `utils`, used by the subset, sw-cache and multifile indexes. Still deferred: the two subset directory naming schemes, since renaming strands the subset dirs already on disk.
8. `makelink` no longer raises on every dry run.

Not fixed here, and worth its own issue: `subset-partition` on heavy-only input fails because of the completion test, not the pairing info. It expects the top-level paired `partition-igh.yaml`, which a heavy-only run never writes, so `finish_process` sees exit 0 with a missing output. Plain `partition --paired-loci` on the same input exits 0 and writes `single-chain/partition-igh.yaml`, and `--no-pairing-info` makes no difference. The fix is to accept the single-chain output and revive your commented-out `keep_all_unpaired_seqs` merge branch, both in the paired merge path this PR does not touch.

PR #409, opened 2026-09-16
Stream Zig HMM input line-by-line, raise cap to 4 GiB, and improve FileTooBig diagnostics

### LINKED ISSUE #408: zig ham backend caps each per-proc hmm input at 64 MiB, which puts a floor under --n-procs
The zig ham backend reads each per-proc hmm input in one allocation with a hardcoded 64 MiB ceiling, at [`packages/zig-core/src/ham/args.zig#L152`](https://github.com/psathyrella/partis/blob/main/packages/zig-core/src/ham/args.zig#L152):

```zig
const content = try std.fs.cwd().readFileAlloc(allocator, self.infile, 64 * 1024 * 1024);
```

The C++ ham it replaces streams the same file line by line with `ifstream` plus `getline`, at [`src/args.cc#L101-L115`](https://github.com/psathyrella/ham/blob/f48ee29ee86e645b202c096d38466b7bd349142e/src/args.cc#L101-L115) in the `packages/ham` submodule, pinned at `f48ee29`, so it has no equivalent limit.

This is a zig-based effect, not a python one. Partis is doing the ordinary thing on its side: it splits the hmm input evenly across `--n-procs`, so a lower proc count makes each per-proc file larger. The zig read is what turns a large file into a hard failure, and the C++ path accepts the same file. The one python-side point, under Error reporting below, is how the failure is surfaced rather than why it happens.

The consequence is that `--n-procs` has a floor as well as a ceiling. Past the floor a run crashes rather than running slower, which inverts the usual expectation.

This was already anticipated in #342, which carries the note from the #351 review that "`args.zig:152` has a 64 MiB cap on `readFileAlloc(self.infile, ...)`. Could become the next `FileTooBig` on a larger run. Bump alongside the next memory-related item, or file separately." This is that separate filing, with the run that hit it, and it is filed at @matsen's request on #407.

#### Measured

Soto Donor03, igh, one 500k subset. Array 8235843, 2026-09-15.

| `--n-procs` | largest per-proc `hmm_input.csv` | rows per proc | fraction of the 64 MiB cap | result |
|---|---|---|---|---|
| 4 | 70,788,907 B | 125,000 | 105.5% | `error: FileTooBig` |
| 8 | 35,401,701 B | 62,500 | 52.8% | ok |
| 16 | 17,708,052 B | 31,250 | 26.4% | ok |

The unsplit input is 283,105,200 bytes over 500,000 rows, so the row width is 566.2 bytes, and the split files agree at 566.3. Against the cap of 67,108,864 bytes that puts the per-proc ceiling at 118,522 rows, so the floor for a 500k subset is 5 procs, and the 4 proc arm misses by 3.68 MB, 5.5%.

Row width is dominated by the colon-joined `only_genes` list, for example:

```
Donor03_chunk_0218_000000021703727 305 306 16 25 0.030470914127423823 45 IGHD3-22*01:IGHD3-22*01_a14t_c28-:IGHD3-22*01_a14t_c28t_t29a_a30c:IGHD3-22*01_c28-:IGHJ4*02:IGHV3-23*01 NNNNNNNNNGAGG...
```

That list grows with the number of genes the glfo keeps, so the row width, and therefore the floor, are dataset-dependent. Neither should be turned into a fixed sequence count.

#### Where this was observed

First seen on 2026-09-14, array 8158769 task 2 (job 8158772), while measuring a proc-count curve. Python side at commit `7d440841e` with an uncommitted feature patch, which is why the numbers above come from the clean rerun instead.

The numbers and the reproducer below are from array 8235843 on 2026-09-15, run from `/fh/fast/matsen_e/scanchi/partis` at commit `9f3240c98`, tree clean apart from untracked zig build artifacts. `packages/zig-core` is byte-identical between that branch, `dev`, and `main`, and `args.zig:152` has been unchanged since `bd769dfb8` (2026-05-09, #377 and #378), so the cap is not specific to any build or branch.

The original crash, quoted from the 2026-09-14 job:

```
Exception: exceeded max number of tries (1 >= 1) for subprocess with command:
        .../packages/zig-core/zig-out/bin/partis-zig-core --algorithm viterbi --hmmdir .../parameters/igh/sw/hmms --datadir /tmp/scanchi/hmms/<hash>/germline-sets --infile /tmp/scanchi/hmms/<hash>/hmm-0/hmm_input.csv --outfile /tmp/scanchi/hmms/<hash>/hmm-0/hmm_output.csv --locus igh --random-seed 1 --only-cache-new-vals --ambig-base N
        stderr:           /tmp/scanchi/hmms/<hash>/hmm-0/err
            error: FileTooBig
```

#### Reproducer

Data, all readable under `matsen_e`:

| path | contents |
|---|---|
| `/fh/fast/matsen_e/processed-data/partis/soto-bcr/sw-annot-all/subsets/Donor03_subset_4M.fasta` | the 4M source this 500k subset came from |
| `/fh/fast/matsen_e/scanchi/tmp/023-nprocs-curve/nprocs-4/igh.fa` | the input used below, 201,406,510 bytes, 500,000 igh sequences |
| `/fh/fast/matsen_e/scanchi/tmp/023-nprocs-curve/nprocs-8/parameters/igh` | completed parameters for that input, including `sw-cache.yaml` at 1,197,482,728 bytes |
| `/fh/fast/matsen_e/scanchi/tmp/023-zig-hmm-input-size` | the 2026-09-15 run: script, logs, per-proc-count workdirs |

The cheap form, about 5 minutes per arm. `annotate` with the cached Smith-Waterman results reaches the hmm step without recomputing anything, and `--workdir` on shared storage keeps the per-proc files where they can be measured, unlike the default under `/tmp` on the compute node:

```bash
PARTIS=/fh/fast/matsen_e/scanchi/partis           # commit 9f3240c98
CURVE=/fh/fast/matsen_e/scanchi/tmp/023-nprocs-curve
NP=4                                              # 4 fails, 8 and 16 succeed

$PARTIS/bin/partis annotate --locus igh \
  --infname $CURVE/nprocs-4/igh.fa \
  --parameter-dir $CURVE/nprocs-8/parameters/igh \
  --sw-cachefname $CURVE/nprocs-8/parameters/igh/sw-cache.yaml \
  --n-procs $NP --random-seed 1 --no-time-based-n-proc-reduction --dont-write-git-info --zig \
  --workdir /tmp/zig-cap-$NP --outfname /tmp/zig-cap-$NP/annotations.yaml

# the per-proc inputs the backend is about to read
find /tmp/zig-cap-$NP -name hmm_input.csv -printf '%s %p\n' | sort -n
```

`--workdir` must not exist beforehand, since partis rejects a pre-existing one.

The original form, about 75 minutes, is `cache-parameters` rather than `annotate`, which recomputes Smith-Waterman:

```bash
$PARTIS/bin/partis cache-parameters --paired-loci --no-pairing-info \
  --infname /fh/fast/matsen_e/scanchi/tmp/023-nprocs-curve/input-500k.fa \
  --paired-outdir /tmp/zig-cap-cache-params \
  --n-procs 4 --random-seed 1 --no-time-based-n-proc-reduction --dont-write-git-info --zig
```

#### Environment

The cap itself is a source constant and does not depend on the environment. These details are here so the run can be repeated exactly.

- Fred Hutch cluster, `campus-new` partition, 16 cpus and 48 GB per task.
- Partis invoked from a pixi environment, `eval "$(pixi shell-hook)"` in `/fh/fast/matsen_e/scanchi/partis-projects-dev`, with `PYTHONPATH` pointed at the partis checkout.
- zig 0.15.2. The binary was built with the two-argument form of [`bin/zig-build.sh`](https://github.com/psathyrella/partis/blob/main/bin/zig-build.sh), target `x86_64-linux-gnu.2.27` plus a search prefix holding `libz`, from a deactivated shell:

  ```bash
  env -i HOME="$HOME" PATH=/usr/local/bin:/usr/bin:/bin bash bin/zig-build.sh \
      x86_64-linux-gnu.2.27 /fh/fast/matsen_e/scanchi/tmp/zig020-prefix
  ```

The positional arguments come from the version of that script on the feature branch, added in #402. On `main` the same build is `zig build -Doptimize=ReleaseFast -Dtarget=x86_64-linux-gnu.2.27 --search-prefix <prefix>` inside `packages/zig-core`. The explicit target is needed because the build host runs glibc 2.39 and the compute nodes run 2.27, so a native build loads on the build host and then fails on a node with `GLIBC_2.34 not found`. An explicit `-Dtarget` also stops zig searching native paths, which is why `libz` needs the search prefix.

#### A question about the caps

Were these picked from measurements? The ceilings differ across call sites, and I do not know what sets any of them:

| call site | cap, binary units |
|---|---|
| `ham/bcr_utils/germ_lines.zig:125` | 4 MiB |
| `ham/hmm_yaml.zig:90` | 16 MiB |
| `ham/args.zig:152` | 64 MiB |
| `ham/bcr_utils/germ_lines.zig:80` | 64 MiB |
| `ham/glomerator/glomerator.zig:511` | 4 GiB |

Separately, and independent of the ceiling: the zig hmm step holds the whole per-proc input resident where the C++ path did not, so peak memory scales as input divided by proc count. That follows from reading the whole file rather than from the ceiling.

#### Error reporting

Whatever the ceiling ends up being, an input can still exceed it, and today that case is hard to diagnose. `error: FileTooBig` reaches the user as `exceeded max number of tries`, which names neither the file, nor its size, nor the fact that a limit was hit at all, so the obvious reading is that the backend crashed for some unrelated reason.

Two places this could be improved, and I do not know which you prefer:

- partis could catch this failure from the backend and say that the per-proc hmm input exceeded what the backend can read, name the file and its size, and suggest a higher `--n-procs`. It knows the size of every per-proc file it wrote. It should not name the limit itself, since hardcoding a copy of a value that lives in the zig source is how the two drift apart.
- the zig read could report the size it saw and the ceiling it was compared against, which is the only place both numbers are known without duplicating one of them.

#### Where the PR should go

`packages/zig-core` is byte-identical on `main`, `dev`, and the feature branch, so a fix belongs on `main` and merges down without conflict. The feature line picks it up at its next merge from `main`, and nothing needs backporting.

matsen:
Closes #408

### Summary

In #408, running with `--zig` and `--n-procs 4` on a 500k-sequence subset failed with `error: FileTooBig`. In `packages/zig-core/src/ham/args.zig`, `readInfile` used `readFileAlloc` with an arbitrary cap of 64 MiB (`64 * 1024 * 1024`). Because the 4-proc per-proc `hmm_input.csv` was 70.8 MB (105.5% of cap), the run crashed with `error: FileTooBig`, surfaced to the user as `exceeded max number of tries`.

This PR:
1. **Replaces `readFileAlloc` with a dynamically-growing streaming line reader**:
   - Instead of allocating the entire raw file into memory at once, `args.readInfile` streams lines via `file.readerStreaming` starting with an initial 2 MiB buffer.
   - If a line (e.g. an oversized cluster row during partition mode) exceeds the buffer, it dynamically grows (doubling up to `max_infile_bytes`), completely avoiding arbitrary fixed line length limits (`StreamTooLong`) while keeping memory minimal for typical single-sequence rows.
   - Raises the default maximum file ceiling `max_infile_bytes` from 64 MiB to 4 GiB (checked via `stat()` up front).
   - Reduces transient memory during input ingestion (overall peak RSS of the process remains dominated by downstream HMM decoding).
2. **Eliminates intermediate allocations & tightens memory lifecycle**:
   - Replaced `ham_text.splitString` on colon-separated fields with zero-allocation `std.mem.splitScalar`.
   - Added `errdefer` cleanup for owned header strings and parsed query collections to guarantee clean rollback and prevent memory leaks on OOM.
3. **Adds actionable error reporting & shared diagnostics**:
   - Centralized file size error reporting into `ham_text.printFileTooBig` across both `args.zig` and `glomerator.zig`.
   - Propagates `stat()` errors explicitly.
4. **Adds unit tests** in `args.zig`:
   - All tests use `std.testing.tmpDir` to avoid `/tmp` races on shared build hosts.
   - Added tests for dynamic buffer growth on lines exceeding initial buffer capacity, unterminated final rows, and rows crossing buffer refill boundaries.
5. **Improves Python error diagnostics and retry handling**:
   - In `partis/utils.py:finish_process`, detects both `FileTooBig` and `StreamTooLong` in subprocess stderr (bounding stderr reads to 1 MiB).
   - Prevents futile retries of identical oversized files and clarifies error messaging (distinguishing fatal non-retryable errors from exhausted retry counts).
   - Uses safe CLI argument extraction (`extract_cli_arg`).
   - Gates the suggestion to increase `--n-procs` so it only fires when `--infile` exceeds `FileTooBig` (suppressed for `--input-cachefname` and `StreamTooLong`, where adding procs does not reduce row/file size).

### Verification

- Rebuilt `partis-zig-core` with `-Doptimize=ReleaseFast`.
- `packages/zig-core`: `zig build test` passes all tests (including dynamic buffer growth, refill boundaries, unterminated lines, and `FileTooBig`).
- Python test suite: `test.py --quick --zig` and `test.py --no-tree-gen --no-simu --zig` pass cleanly with all reference comparisons matching.


s-canchi (commented):
Thank you for turning this around so quickly. I built the branch at `9fd29b8af` on the cluster. I ran it against the data from #408, and against the light-chain sets I use for scale work.

Two things came out of that. The `FileTooBig` fix works, and the parser rewrite produces byte-identical output. But the fixed 2 MiB line buffer is a regression. It breaks the HA re-partition stage on real data, so I think that one change needs rework before merge.

Every comparison below is the pre-PR binary against the branch binary. Same inputs, same flags. The pre-PR binary was built on 2026-09-11, and `packages/zig-core` is byte-identical across `main`, `dev`, and my feature branch.

#### The 2 MiB line buffer is a regression, with no `--n-procs` escape

`readInfile` now streams with a fixed `read_buffer_size` of 2 MiB. Any row longer than that fails with `StreamTooLong`. The old `readFileAlloc` had no per-line bound at all, only the whole-file one.

This matters because in `partition` mode a row is a whole cluster, not a single sequence:

- [`write_to_single_input_file`](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L2204-L2212) writes `names` and `seqs` as colon-joined member lists.
- The subcluster split that keeps annotation rows small is gated on `algorithm == 'viterbi'` at [`partitiondriver.py#L1652`](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L1652). It does not apply to the forward and agglomeration rounds.

Measured on one real light-chain cluster of 19,112 sequences (1M simulation, donor 1, igk, cdr3-33, sub-000):

| binary | result | longest row read |
|---|---|---|
| pre-PR | exit 0, completes | 3,679,856 B (3.51 MiB) |
| this branch | exit 1, `error: StreamTooLong` | fails at 2,846,475 B |

It fails on the first agglomeration step, `istep-0/hmm-0`, not after several rounds. The naive-collapsed initial partition already holds the oversized row. Backend stderr:

```
error: data row in '.../istep-0/hmm-0/hmm_input.csv' exceeds maximum supported line length (2 MiB)
error: StreamTooLong
```

Raising `--n-procs` does not help. [`split_input`](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L1874-L1890) reads with `csv.DictReader` and hands out whole rows, so one row is never split across procs. This is the part I would flag hardest. #408 was a whole-file ceiling, and `--n-procs` was a real workaround for it. This replaces it with a per-line ceiling that has no workaround.

How much of the work here that touches, using measured field widths (uid 34 B and padded sequence 368 B on igh, uid 24 B and 340 B on igk):

- A cluster overflows the buffer above roughly 5,200 members on heavy, and above roughly 5,700 on light.
- All six light 1M cells have pre-refine clusters above that. The largest is 20,056 members.
- On the 4M heavy donor, the final partition holds nine clusters between 10,001 and 100,000 members, and a maximum of 44,194. Clusters going into HA are at least as large as the ones coming out, so heavy is exposed too. I have not measured a heavy pre-HA row directly, so that last point is inference rather than measurement.

Reproducer, all readable under `matsen_e`:

```bash
IN=/fh/fast/matsen_e/scanchi/tmp/014-cache-d1-1m/groups-rerun/single-chain/disjoint-groups/igk/\
groups/cdr3-33/sub-groups/sub-000/ha-repartition/clusters/cluster-000003-size-19112/input.fa

partis partition --locus igk --infname $IN \
  --parameter-dir <igk params> --sw-cachefname <group sw-cache> \
  --no-indels --no-naive-vsearch --n-procs 8 --random-seed 1 \
  --no-time-based-n-proc-reduction --dont-write-git-info --zig
```

Full logs: `/fh/fast/matsen_e/scanchi/tmp/023-zig-409-verify/ha-8356145.out`.

On the fix. I do not think a larger constant is the answer. Any fixed number brings the same failure back at a different size. `readInfile` already calls `stat()` on the file, so the buffer could be sized from that. Growing the buffer on `StreamTooLong` instead of failing would also work. Either way the hardcoded number leaves the hot path, rather than 64 MiB being traded for 2 MiB.

#### The new diagnostics do not cover the failure this PR introduces

In the run above, the user saw the old generic message:

```
Exception: exceeded max number of tries (1 >= 1) for subprocess with command: ...
```

`finish_process` looks for the string `FileTooBig` in stderr. This failure is `StreamTooLong`, so none of the new diagnosis fired. That run had `n_max_tries` of 1. Under a batch system it is 3, so it would also retry twice for something that cannot succeed. It would be worth keying the detection on both error names, or on whatever covers every non-retryable read failure.

Two smaller notes on the same code:

- The hint always recommends raising `--n-procs`, including when the oversized file came from `--input-cachefname`. That file is copied whole to every subdir at [`partitiondriver.py#L1907`](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L1907). More procs makes more copies of the same size, not smaller ones. The hint could be gated on the `--infile` match.
- `extract_cli_arg` catches `IndexError`. The bounds check above it already makes that unreachable.

#### The peak-RSS claim is not supported by measurement

The PR body says the streaming reader drops peak RSS during input loading. Transient RSS during parsing does drop. Peak RSS does not change, because the peak happens during HMM decoding, and the whole-file buffer is freed before then.

Same 35.4 MB per-proc input, same flags:

| binary | peak RSS |
|---|---|
| pre-PR | 254,492 KB |
| this branch | 254,592 KB |

The branch is 100 KB higher, not one input file lower.

For scale planning, the useful number is different. RSS runs at roughly 2.2 KB per query on top of a 116 MB floor, measured at 62,500 and at 125,000 queries per proc. So the new 4 GiB ceiling, about 7.5 million rows per proc, would need roughly 17 GB in one process. The ceiling is now unreachable before memory is. That is the right direction. It does mean 7.5 million rows per proc is not a configuration anyone can actually run, so it may be worth not describing it as accommodated.

#### What I verified as working

| check | result |
|---|---|
| the #408 failure, `annotate` at `--n-procs 4` on the 500k subset | completes, 500,000 hmm output lines |
| pre-PR binary on the same input, as a control | `error: FileTooBig` in 0.8 s, as reported |
| pre-PR against branch, 500k end to end, `--n-procs 8`, `PYTHONHASHSEED=0` | byte-identical output |
| pre-PR against branch, same real 35.4 MB per-proc input, backend only | byte-identical, 43,733,771 B |
| input with the final newline stripped, at 35 MB and at 200 rows, both binaries | byte-identical output, no row lost |
| the new 4 GiB message | names the path, the actual size, and the cap, and exits through `stat()` in 0.01 s |
| whether `error: FileTooBig` still reaches stderr next to the new message | it does, so the python detection keys off something real |
| `finish_process` on a `FileTooBig` stderr | no retry, names the file and its size, as #408 asked |
| `finish_process` on an ordinary failure | still retries, and the exhausted-retries message is unchanged |
| `zig build test` | passes |

#### On the caps question from #408

I measured the three untouched ceilings against the largest real files in the parameter sets here. That question can be closed as not urgent, whatever the original reasoning was:

| call site | cap | largest real file | headroom |
|---|---|---|---|
| `germ_lines.zig:80` | 64 MiB | 17,170 B (`ighv.fasta`) | 3,900x |
| `germ_lines.zig:125` | 4 MiB | 1,459 B (`extras.csv`) | 2,900x |
| `hmm_yaml.zig:90` | 16 MiB | 119,602 B (per-gene hmm) | 140x |

One related note. `glomerator.zig` still reads the partition cache with whole-file `readFileAlloc` at 4 GiB, while `args.zig` moved to streaming. That is out of scope here. It is worth flagging because the cache grows across agglomeration rounds, so it is the one place where the 4 GiB ceiling still means 4 GiB resident.

#### Suggested tests

The current test covers the `stat()` rejection path with a 20 byte ceiling. The rewrite's real risks are elsewhere, and two of the three below are where I expected it to break:

- a row longer than the read buffer, which is the failure above;
- an input whose last row has no trailing newline, which I confirmed works but which nothing pins;
- a row that spans a buffer refill boundary, which the 70 MB real input covers by accident but no test asserts.

#### Smaller items

- The cumulative `total_bytes_read` check cannot fire for a regular file. The `stat()` check above it already bounds the total. It also adds a per-row branch, and counts a trailing byte for an unterminated final line.
- The `FileTooBig` message is built four times across `args.zig` and `glomerator.zig`. Each copy has its own branch for whether the cap is at least 1 GiB. One helper would do.
- `if (file.stat()) |st| ... else |_| {}` silently skips the size check when `stat` fails.
- Two comments were dropped from `readInfile`. One recorded that the colon split matches C++ `SplitString(tmpstr, ":")`. The other was on the sequence loop. The first seems worth keeping in a file whose header is about C++ correspondence. The `splitScalar` substitution itself is correct, including the empty-field case.
- `const ham_text` in `args.zig` is now unused in that file.
- The integer-header chain is reformatted onto one very long line. That is harder to read than the wrapped original, and it is unrelated to the fix.
- The new test writes to a fixed absolute path under `/tmp`. Two runs of the suite on one host race on it. On a shared build host the file can be owned by another user, so the create fails with a permission error rather than a useful test failure. `std.testing.tmpDir` avoids both. The existing test in the file does the same thing, so this is convention rather than something this PR introduced.

matsen:
Thank you @s-canchi for the incredibly thorough review, the scale measurements, and for identifying the large-cluster row length regression!

All review items have been addressed in commit `d98f178b2`:

1. **Dynamic line buffer growth**:
   - `readNextLine` now catches `error.StreamTooLong` and dynamically doubles the read buffer (starting from 2 MiB up to `max_infile_bytes`), preserving unconsumed bytes and retrying seamlessly.
   - This eliminates the hardcoded per-line cap for large clusters in partition mode (including the 19k+ sequence clusters in HA re-partition) while keeping memory minimal for typical single-sequence rows.

2. **Python error handling & diagnostics polish (`partis/utils.py`)**:
   - `finish_process` now recognizes `StreamTooLong` alongside `FileTooBig` as fatal non-retryable errors, preventing futile retries.
   - The suggestion to increase `--n-procs` is now gated so it only appears when an oversized file is `--infile` with `FileTooBig` (suppressed for `--input-cachefname` and `StreamTooLong`, where adding processes does not split clusters or reduce cache size).
   - Removed unreachable `IndexError` in `extract_cli_arg`.

3. **Size checks, diagnostics consolidation & cleanup**:
   - Size check is performed via `file.stat()` up front and propagates stat errors rather than silently skipping them.
   - Removed the redundant per-row `total_bytes_read` counter and branch.
   - Added a centralized `ham_text.printFileTooBig` helper in `text.zig`, eliminating duplicated GiB-branching and formatting across `args.zig` and `glomerator.zig`.
   - Restored dropped comments regarding C++ `SplitString` and sequence loop, and restored multi-line wrapped formatting for integer headers.

4. **Testing**:
   - Migrated existing tests in `args.zig` to `std.testing.tmpDir` to eliminate shared-host `/tmp` races and permission conflicts.
   - Added unit tests explicitly asserting:
     - Dynamic buffer growth when encountering a line longer than the initial buffer.
     - Files with an unterminated final row (no trailing newline).
     - Rows that cross buffer refill boundaries.

5. **PR Description**:
   - Updated the PR summary to reflect the dynamic buffer growth mechanism and clarified that transient RSS during ingestion is reduced, while overall process peak RSS is dominated by downstream HMM decoding.

matsen:
### Verification against Saranya's exact datasets

I verified the updated binary directly against the files in `/fh/fast/matsen_e/scanchi/tmp/023-zig-409-verify/`:

1. **HA re-partition 19,112-member cluster (`istep-0/hmm-0/hmm_input.csv`, 2,846,475 byte row)**:
   - Command:
     ```bash
     partis-zig-core --algorithm forward        --hmmdir .../parameters/igk/hmm/hmms        --datadir .../ha-work-pr/germline-sets        --infile .../ha-work-pr/istep-0/hmm-0/hmm_input.csv        --input-cachefname .../ha-work-pr/istep-0/hmm-0/hmm_cached_info.csv        --partition --locus igk ...
     ```
   - **Result**: Exit code **0** in 1.5s (previously failed with `error: StreamTooLong` on the 2.71 MiB row). Processed all 19,112 naive sequences with dynamic buffer growth.

2. **#408 70.8 MB per-proc input (`work-4/hmm-1/hmm_input.csv`, 125,000 queries)**:
   - Command:
     ```bash
     partis-zig-core --algorithm viterbi        --hmmdir .../parameters/igh/hmm/hmms        --datadir .../work-8/germline-sets        --infile .../work-4/hmm-1/hmm_input.csv        --outfile .../A1-test.out.csv --locus igh ...
     ```
   - **Result**: Exit code **0** in 206.7s (`calcd: vtb 125000`).
   - Diff against Saranya's `work/A1-new.out.csv`: **byte-for-byte identical (0 differences across 84 MB / 125,001 rows)**.

s-canchi (commented):
Thank you for the quick turnaround on this. I rebuilt `d98f178b2` on the cluster and re-ran the failing case. The line buffer regression is fixed, and the fix is correct on real data, not only non-crashing.

Your HA check reports exit code 0, which is the right first signal, but exit 0 alone does not show that the grown buffer parsed the oversized row correctly. A bad copy on the growth path would give a mangled name or sequence list rather than a crash. The #408 70.8 MB input does not cover that either, because its longest row is well under 2 MiB, so it never triggers growth. So I ran the byte comparison as well.

#### Confirmed fixed

Direct backend calls, both binaries, on the preserved `istep-0/hmm-0/hmm_input.csv` with the 2,846,475 byte row:

| binary | result | output | output cache |
|---|---|---|---|
| pre-PR (2026-09-11) | exit 0, 0.41 s | 1,972,068 B | 1,191,003 B |
| `d98f178b2` | exit 0, 0.29 s | 1,972,068 B | 1,191,003 B |

Both files byte-identical between the two binaries.

Full HA re-partition on the same 19,112-member igk cluster, `PYTHONHASHSEED=0`, `--n-procs 8`:

| check | result |
|---|---|
| pre-PR and `d98f178b2` | both exit 0 |
| longest row actually read | 3,679,856 B on both, so growth past the 2 MiB initial buffer is exercised |
| final annotations | byte-identical, 13,975,160 B |
| `zig build test` | passes, including the three new tests |

Log: `/fh/fast/matsen_e/scanchi/tmp/023-zig-409-verify/ha2-8358698.out`.

#### One item still open: the `--n-procs` gate

This one is on my earlier wording. I wrote "gated on the `--infile` match", which reads as "check that an `--infile` is present". What I meant was "check that the file which overflowed is the `--infile`". The current code implements the first reading, so the gate never closes:

`has_oversized_infile` is set whenever the `--infile` path exists on disk, which is always true. [`get_hmm_cmd_str`](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L1289) always passes `--infile`, and [line 1302](https://github.com/psathyrella/partis/blob/main/partis/partitiondriver.py#L1302) adds `--input-cachefname` on top in partition mode, so the two always coexist. A `FileTooBig` raised on the cache file therefore still prints the suggestion:

```
error diagnosis: backend failed because an input file or row exceeded the maximum supported size.
  per-proc input file (.../hmm_input.csv): 3145728 bytes (3.0 MB)
  input cache file (.../hmm_cached_info.csv): 5242880 bytes (5.0 MB)
  suggestion: increase --n-procs to split queries into smaller per-proc files.
```

To be exact this time, the rule I am asking for is: print the suggestion only when the path named in the backend diagnostic is equal to the `--infile` value. Your new `ham_text.printFileTooBig` makes that reliable, because every one of those messages now has the same shape, `error: <description> '<path>' ... exceeds maximum supported size`. Concretely, in place of the `has_oversized_infile` flag:

```python
oversized_paths = re.findall(r"error: [a-z ]+ '([^']+)' .*exceeds maximum supported size", err_content)
infile_path = extract_cli_arg(cmdfo.get('cmd_str'), '--infile')
if is_file_too_big and infile_path is not None and infile_path in oversized_paths:
    hint_lines.append('      suggestion: increase --n-procs to split queries into smaller per-proc files.\n')
else:
    hint_lines.append('')
```

If the message shape ever changes, the list comes back empty and the suggestion is dropped, which is the safe direction to fail in.

`StreamTooLong`, `FileTooBig` on `--infile` under `annotate`, and ordinary retryable failures all behave correctly already.

#### Two smaller things

1. The exhausted-growth message reports the buffer length where a line length is expected:

   ```zig
   std.debug.print("error: line in '{s}' ({d} bytes) exceeds maximum supported line length ({d} bytes)\n",
                   .{ filename, buffer.*.len, max_line_bytes });
   ```

   Growth clamps the new size to `max_line_bytes`, so at the point of failure `buffer.len` is exactly `max_line_bytes`. Both numbers print as 4294967296, and the first one is not the length of the offending line, which the reader never learns. The true length is not known at that point, so I would just drop it:

   ```zig
   std.debug.print("error: line in '{s}' exceeds maximum supported line length ({d} bytes)\n",
                   .{ filename, max_line_bytes });
   ```

2. `read_buffer_size` became `pub var initial_read_buffer_size`, a mutable process-global, so that the two new tests can shrink it to 64 bytes. A field on `Args` gives the tests the same control without process-global mutable state: declare `initial_read_buffer_size: usize = 2 * 1024 * 1024` on the struct (setting it in `initDefaults` as well if that function fills fields one by one), use `self.initial_read_buffer_size` in `readInfile`, and have the tests set `args.initial_read_buffer_size = 64` after `initDefaults` instead of saving and restoring a global.

Optional, unrelated to correctness: growth is bounded by `max_infile_bytes` rather than by the size `stat()` already returned, so a malformed 4 GiB file containing no newline would grow the buffer to 4 GiB, which is the whole-file load this PR set out to avoid. Passing `@min(self.max_infile_bytes, file_stat.size + 1)` as `max_line_bytes` changes nothing on real inputs.

None of the three block the merge from my side. The regression that did is fixed. The `--n-procs` gate is the one I would still like corrected, since wrong advice on a fatal error costs someone a wasted rerun, but it is diagnostic text and not data.

s-canchi (approved):
LGTM!

matsen:
Thanks @s-canchi! All four follow-up items have been implemented in `1681570cb`:

1. **`--n-procs` gate in `partis/utils.py`**: Added module-level `import re` and gated the recommendation with regex path matching so `--n-procs` is only suggested when the file reported as overflowing in stderr matches the `--infile` argument.
2. **Buffer size configuration on `Args`**: Removed the mutable global `pub var initial_read_buffer_size`, added `initial_read_buffer_size` to `Args` (defaulting to 2 MiB), and updated tests to configure this field per instance.
3. **Diagnostic message formatting**: Removed the redundant `({d} bytes)` buffer length from `readNextLine` error reporting.
4. **Line-length bounding**: Passed `max_line_bytes = @min(self.max_infile_bytes, file_stat.size + 1)` in `readInfile` to prevent buffer growth beyond actual file size.

matsen:
If you're satisfied now, let Duncan know and we can get this merged in.

psathyrella:
> Claude Code:

Built the pinned zig 0.15.2 (per `bin/zig-build.sh`) and ran `zig build test` in `packages/zig-core`: 67/67 pass. The findings below are checked against the real std 0.15.2 source and, where noted, against the compiled binary.

## Verified correct

`takeDelimiter` **excludes** the delimiter and returns `null` (not an error) at true EOF -- `lib/std/Io/Reader.zig:838-851` in the pinned toolchain. So the trailing field of each row is clean and `parseFloat` on the last column is not at risk, and the "input without trailing newline" test exercises a real path.

The direct writes to `file_reader.interface.buffer` / `.seek` / `.end` (args.zig:160-184) are supported rather than a hack: the VTable doc at `Reader.zig:73-77` says implementations may write directly to `buffer` and modify `seek` and `end`, and `File.Reader.logicalPos` (`File.zig:1290`) depends only on `end - seek`, which the compaction preserves exactly. File position tracking stays correct.

No leaks and no dangling slices. The `errdefer q.deinit(allocator)` at args.zig:203 fires on any mid-row `try` failure and does not fire after a successful `append`, so no double-free -- checked by running a row with a non-numeric `k_v_min` under `std.testing.allocator`, no leak reported. The per-field `errdefer allocator.free(owned)` statements are not dead code: each guards the `append()` that follows a successful `dupe()`. Every slice `readNextLine` returns into the reader's buffer is fully consumed before the next call that could reallocate it. CRLF is handled, by the same `std.mem.trim(u8, raw_line, " \t\r")` as before.

## One bug

**`readNextLine` loops forever if `initial_read_buffer_size == 0`** (args.zig:171-174). `new_size = buffer.*.len * 2` stays 0, so each iteration allocates a zero-length buffer, frees the old one, and takes `StreamTooLong` again. Reproduced: with `args.initial_read_buffer_size = 0` and a two-line CSV the process hung and had to be killed on timeout.

Not reachable today, since `initial_read_buffer_size` and `max_infile_bytes` have no CLI wiring and production always gets the 2 MiB default. But it is a public field with no validation, and two of the new tests already set it (`= 64`), so a dropped zero hangs rather than failing. `@max(buffer.*.len * 2, 1)`, or return an error when growth does not increase capacity.

## Tests that do not assert anything

The three `printFileTooBig(...)` calls in text.zig:220-224 can only fail by panicking, and there is no panic path: the format strings are comptime-checked whether or not the test runs, and `std.debug.print` returns nothing. The doc comment above `printFileTooBig` states an exact output format and nothing checks it. Building the string with `std.fmt.allocPrint` and printing that would make the format testable.

## Coverage gaps, in the order I would worry about them

1. `initial_read_buffer_size = 0` -- would have caught the hang above.
2. **No row of realistic shape.** The dynamic-growth test uses `n1` / `s1` / `gene1` tokens totaling a bit over 200 bytes. The row that produced #408 is 566.3 bytes, dominated by a colon-joined `only_genes` of full gene names. Nothing exercises the streaming parse on a row of that shape, and nothing compares the resulting `queries` against an expected structure taken from a real `hmm_input.csv`.
3. No empty-file test (`EmptyInputFile`) and no header-only test (zero rows, no error). Both are old behavior, but the read path underneath them was replaced.
4. No CRLF test. Correct today, traced by hand, and exactly the kind of thing a later refactor breaks silently.
5. No mid-row parse-failure test. The cleanup this PR adds does work, but only because it was checked by hand.

## Python side

`import re` is genuinely needed: `re` was not imported in utils.py, and the only prior `re.` on main is inside a commented-out assert at utils.py:4473.

The stderr regex matches what the zig prints and correctly does not match the line-length message. Ran all three forms against `r"error: [a-z ]+ '([^']+)' .*exceeds maximum supported size"`: `input file` and `cache file` match and capture the path, `error: line in '...' exceeds maximum supported line length` does not. The `--n-procs` suggestion stays gated to `--infile` FileTooBig as described.

**`'FileTooBig' in err_content` works, but not for the reason it looks like.** `printFileTooBig` never emits the literal string `FileTooBig`; detection depends entirely on zig's top-level `error: <name>` line. That still happens -- `main() !void` does `try bcrham.run(...)` (main.zig:28), errors propagate out, and the stderr captured in #408 shows `error: FileTooBig` with no stack trace, which is the ReleaseFast shape. It is an undocumented coupling between the Python matcher and zig's default error printing, so a `catch` added later in `main` or `bcrham.run` would silently disable the whole diagnosis path. Worth a comment at the Python end saying what it depends on.

`ef.read(1024 * 1024)` (utils.py:5899) reads the **first** 1 MiB of stderr. If the backend wrote more than that before dying, the `error:` line is at the end and detection silently reverts to the old behavior. The tail, or the whole file, is the right end -- this is one backend process's stderr, so it is already bounded.

**The "increase --n-procs" suggestion is right for `annotate` and misleading for `partition`.** During partition, partis reduces `n_procs` itself across hieraglom iterations (partitiondriver.py:785, `shall_we_reduce_n_procs`), with `--n-procs` acting only as the per-iteration ceiling (:779). Someone who hits this during partition and raises `--n-procs` can watch it get reduced straight back down. Academic at 4 GiB, but the hint is what they will act on.

`extract_cli_arg` (utils.py:5819) splits on whitespace, so it breaks on a path containing a space. Not a real case in partis, but it is a new general-purpose helper in utils.py and will get reused.

The err file is truncated per try (`open(ferr, 'w')`, utils.py:5709), so a stale `FileTooBig` from a previous attempt cannot suppress a legitimate retry.

## Two points about scope and framing

**The cap raise is what closes #408; the streaming is a separate improvement.** The title and summary lead with streaming. The old path died because 70.8 MB exceeded a 64 MiB `readFileAlloc` cap; raising that cap to 4 GiB is the fix. Streaming additionally drops transient ingestion memory by roughly the file size, which against a multi-GB process is real but small -- as the summary's own hedge about peak RSS says. Worth stating which change closes the issue.

**The floor still exists; say where it is now.** At the 566.3 bytes per row measured in #408, a 4 GiB per-proc cap puts it at about 7.6M rows in a single proc. Effectively gone for real data, but it is the same kind of limit, and the next person to hit it will want the number.

> Added `errdefer` cleanup for owned header strings and parsed query collections to guarantee clean rollback and prevent memory leaks on OOM.

This reads as a leak fix and is not one. The old `headers` held slices into `content`, which lived for the whole function; once lines are transient the headers have to be duped, and duping is what creates the obligation to free. It is a consequence of streaming, not an independent improvement.

**`glomerator.zig` is a deferral, not a fix.** Its cap was already 4 GiB (glomerator.zig:511) -- this PR adds only the diagnostic. The partition cache path still does the single whole-file `readFileAlloc` that is the same bug class as #408, with more headroom, and the comment directly above says that cache "grows past 256 MiB" on collision-heavy paired runs. If that is a deliberate scope cut it should say so at the call site, since the new diagnostic makes it read finished.

**Follow-up, not a blocker:** the manual growth loop could be std. `Writer.Allocating` (`Writer.zig:2516`) with `Reader.streamDelimiterLimit` / `streamDelimiterEnding` (`Reader.zig:~930-970`) gives the same grow-with-a-cap behavior and turns overflow into `error.StreamTooLong` for you -- no Reader internals, no doubling loop, and no zero-length case to guard. About 25 lines of index bookkeeping replaced by a couple of calls.

No file overlap with #406, the other open zig-core PR.

s-canchi (commented):
Thank you. All four items land as asked. I rebuilt `1681570cb` on the cluster and re-ran the same checks against that commit.

| check | result |
|---|---|
| direct backend, both binaries, on the 2,846,475 byte row | both exit 0, `hmm_output.csv` and output cache byte-identical (1,972,068 B, 1,191,003 B) |
| full HA re-partition, 19,112-member igk cluster, `PYTHONHASHSEED=0`, `--n-procs 8` | both exit 0, final annotations byte-identical, 13,975,160 B |
| longest row actually read | 3,679,856 B on both arms |
| `zig build test` | passes |

Identical to the numbers at `d98f178b2`, which is the expected result: the new bound changes nothing on real inputs.

The `--n-procs` gate now behaves as intended, against the stderr the current binary actually prints:

| stderr | command shape | suggestion printed |
|---|---|---|
| `error: input file '...' ... exceeds maximum supported size` plus `error: FileTooBig` | `annotate`, the #408 case | yes |
| `error: cache file '...' ...` plus `error: FileTooBig` | `partition` | no |
| `error: line in '...' exceeds maximum supported line length` plus `error: StreamTooLong` | `partition` | no |
| bare `error: FileTooBig` with no descriptive line | `annotate` | no |

The last row is worth naming. The suggestion now depends on the descriptive line being present, so a binary that printed only the bare error name would lose it. That is the safe direction to fail in, and only a pre-PR binary does that. It is the same coupling @psathyrella flagged from the detection end, so a comment in `utils.py` saying that both the detection and the hint key off zig's stderr text would cover the regex as well.

The regression I raised is fixed. Nothing blocking the merge from my side.

PR #410, opened 2026-09-18
ksw, ham: fix three Zig/C++ parity divergences

### LINKED ISSUE #403: Fix three zig/C backend divergences in ksw and fast_math
# Fix three zig/C backend divergences in ksw and fast_math

🤖

Make `partis-zig-igsw` and the C `ig-sw` agree on every input reachable from the partis CLI, and fix an out-of-bounds read in the C++ `bcrham`. One of the three divergences fires on the shipped default human germline set on every zig run; the other two are narrower but reachable.

## Motivation

Found while reviewing PR #402, which forward-ports the already-merged zig work onto the `disjoint-grouping` branch. The zig sources in that PR are byte-identical to `main`, so none of this originates there — these are pre-existing and affect `main` and `dev` today.

This matters more than "opt-in backend" suggests: `bin/partis` defaults to zig on macOS arm64 (issue #330, since arm64 cannot run the SSE2 C `ig-sw`), so Apple Silicon users always run the diverging code path.

PR #402 reports a parity failure on a 30k input: 5 of 6 output files differ, 0.47 to 0.77 percent of sequences in different clusters, 28 flipped queries, with every divergent query differing only in its D gene call. That PR attributes it to `ig_align.c:212` using `ks_introsort` (not a stable sort) with `partis/waterer.py:587` taking the aligner's emission order as the ranking.

Item 1 below is a second, independent divergence producing the same symptom — divergent D calls. It does not explain the tie *rate*, which is set by `ksw_align` scores that match exactly.

The unstable sort cannot produce a divergence on its own. `ig_align.c` is the same C compiled into both backends, and its sort key is `loc.score` alone (`__aln_score_lt`, `ig_align.c:125-126`), which comes from `ksw_align`. Identical `ksw_align` results therefore give identical sort input and identical emission order. So item 1 is the mechanism that has to be ruled out before the introsort explanation can stand.

## Problem

| # | Divergence | Reachable | Measured |
|---|---|---|---|
| 1 | `ksw_global` returns 0 for zero-length target; C computes a real score | Any run using the default human germline set | ~1% of reads on `test/mishmash.fa` |
| 2 | C++ `bcrham`'s `fast_math.c` reads past `softplus_lut` | Any `AddInLogSpace` call with equal or near-equal finite operands | Undefined behavior, currently masked |
| 3 | Lazy-F loop diverges at `gapo == 0` | `--gap-open-penalty 0`, a legal CLI value | 4,830 of 40,000 trials |

Measurements marked below as from the #402 review were produced by randomized differential trials during that review. The harness was never landed, so they are not currently reproducible from the repo; landing it is part of Phase 1.

### 1. `ksw_global` zero-length-target guard

`packages/zig-core/src/igsw/ksw.zig:697` returns early:

```zig
if (qlen_u == 0 or tlen_u == 0) return 0;
```

`ksw.c:474` (`ksw_global`) has no such guard — `:480` zeroes `*n_cigar_` and falls through to allocation. With `ig_align`'s default bandwidth of 150 (`main.zig:80`), longer than any D match, C returns `-(gapo + gape*qlen)` and a full-length I cigar.

The `tlen == 0` case (from `te < tb`) is reached against `IGHD2-8*01_a1t_2t-----------------------------31`, a 2-nt entry at `data/germlines/human/igh/ighd.fasta:91` (sequence `TT`; the next shortest D record is 11 nt). Removing that record from the germline set makes the whole file — all 1,412 reads — byte-identical between backends on `test/mishmash.fa`. The ~1% in the table above is the fraction of those reads that differ before the fix.

This looks like unreachable defensive code until the germline data is checked, which is why it was missed.

### 2. Out-of-bounds read in the C++ `bcrham`'s `fast_math.c`

`fast_math.c` exists as two copies: `packages/zig-core/src/ham/fast_math.c` in this repo, and `src/fast_math.c` in the `packages/ham` submodule (psathyrella/ham). The file's own header says both sides must stay in sync. They are not — `diff` between them is exactly the clamp added by `1625e2efb` ("zig: fix fast_softplus LUT OOB read at d == 0 (#374)") plus its comment, which the ham copy never received:

```c
if (i >= LUT_N) i = LUT_N - 1;
```

Without it, when `AddInLogSpace`'s two operands are equal or near-equal and finite, `t` computes to exactly `LUT_N` (`LUT_SCALE = 65536/30`, so `d = -1e-15` already rounds `30 - d` to `30.0`), giving `i == LUT_N`. The interpolation then reads index `LUT_N + 1` of a table whose valid indices are `0..LUT_N` (`static double softplus_lut[LUT_N + 1];`, `fast_math.c:53`). Two equal `-INFINITY` operands take the early-out at `mathutils.h:106-109` and never reach it.

The read is usually masked because `frac == 0` and the garbage is multiplied by zero, but `0 * NaN` is `NaN`, so masking depends on whatever static data the linker places after the table. It is undefined behavior in the C++ `bcrham`, the default backend everywhere except Apple Silicon, and it would trip ASAN or valgrind.

The clamp cannot change output. For `d <= 0`, `t` never exceeds `LUT_N`, so the clamp fires only where `frac` becomes exactly `1.0`; adjacent LUT entries near `log 2` are within Sterbenz range, so `y1 - y0` is exact and `y0 + 1.0*(y1-y0) == y1` bit-for-bit.

### 3. Lazy-F divergence at `gapo == 0`

`ksw.zig:250-255` (u8 path) and `:385-392` (u16 path) defer the loop break where `ksw.c` uses `goto`:

```zig
f = f -| gape_v; // lazy-F: only decrement, no max with h
```

The u16 path uses `subsEpu16(f, gape_v)` at `:389` for the same step. Provably equivalent for `gapo >= 1`. At `gapo == 0` the two diverge on 4,830 of 40,000 randomized trials (#402 review), with zig returning the strictly higher (correct) score. `--gap-open-penalty` is a plain int defaulting to 30 (`bin/partis:1873`) with no range check anywhere — `partis/processargs.py:334-335` only overrides it under `--no-indels`/`--all-seqs-simultaneous` — so `--gap-open-penalty 0` is accepted and produces backend-dependent results.

## Proposed implementation

Phases are independent and can land separately.

**Phase 1 — item 1.** Delete the `tlen == 0` guard from `ksw.zig:697` so `ksw_global` matches C, and fix the two sites that currently make that unsafe:

- `ksw.zig:787`, `var ti: i32 = @intCast(tlen_u - 1);` — `tlen_u` is `usize`, so at `tlen == 0` this underflows (panic in Debug/ReleaseSafe, wrap then illegal `@intCast` in ReleaseFast). C does the same arithmetic in `int`, gets `-1`, and skips the backtrack loop. Compute as `@as(i32, @intCast(tlen_u)) - 1`.
- `ksw.zig:809`, `var hi: usize = @intCast(n_cigar - 1);` — panics when `n_cigar == 0`, reachable at `tlen == 0` when `w == 0` or `qlen == 0`. C's `for (i = 0; i < n_cigar>>1; ++i)` handles zero naturally. Guard the cigar reversal on `n_cigar > 0`.

Everything else on the `tlen == 0` path is already correct without the guard: allocations are `orelse return 0` and `malloc(0)` is fine, `n_col * 0 == 0`, the DP loop does not execute, and `score = eh[qlen].h` picks up the first-row fill.

Also land the differential harness used for the trial counts above, so those numbers become reproducible.

**Phase 2 — item 2.** Apply the clamp to `src/fast_math.c` in psathyrella/ham, re-pin `packages/ham`, and add a check that fails when the two `fast_math.c` copies differ. The copies are duplicated deliberately for now (the C is expected to go away once zig is trusted), so the check is the fix rather than de-duplication. While there, fix `packages/zig-core/src/ham/fast_math.c:33`, which says "Mirrored on the Zig side at packages/zig-core/src/ham/fast_math.c" — it points at itself, a copy artifact.

**Phase 3 — item 3.** Either reject `--gap-open-penalty 0` in `partis/processargs.py`, or fix the lazy-F loop to match C. If 0 is a legitimate setting, prefer fixing the loop; note that zig is the one computing the correct score, so "match C" would propagate a bug.

## Files to modify

- `packages/zig-core/src/igsw/ksw.zig` — `:697`, `:787`, `:809`; the lazy-F loop at `:250-255` and `:385-392` if Phase 3 takes that route
- psathyrella/ham `src/fast_math.c` — the clamp, then re-pin `packages/ham` here
- `packages/zig-core/src/ham/fast_math.c` — the self-pointing header line at `:33`
- `partis/processargs.py` — `--gap-open-penalty` validation if Phase 3 takes that route
- wherever the copy check lands (a build step in `packages/zig-core/build.zig`, or a test)

## Test plan

1. Both backends on `test/mishmash.fa`, comparing `sw-cache.yaml` per query. Expected after Phase 1: zero divergent queries, matching what removing the 2-nt germline record already produces. This is the reproducible-by-anyone check, since the input is in-repo.
2. `bin/partis-30k-parity-test.sh` before and after Phase 1, recording the divergent-query count both times. Its default input `/tmp/paired-paper-all-seqs.fa` (`:25`) could not be located when PR #402 tried to rerun it; use the substitute that PR named, `vs-shm/v4/seed-0/scratch-mute-freq-0.05`, which is the 30k input that produced the 28 flipped queries.
3. `zig build test` after each phase.
4. Phase 2 only: confirm output is byte-identical before and after the clamp, on any input. The change is value-preserving, so any difference indicates a mistake.
5. Phase 3 only: run both backends with `--gap-open-penalty 0` on a small input and compare.

## Success criteria

1. Both backends produce identical `sw-cache.yaml` on `test/mishmash.fa`, per query.
2. The 30k parity test on the substitute input passes, or its remaining divergences are attributed to a named cause.
3. The two `fast_math.c` copies are identical, and something fails when they are not.
4. `--gap-open-penalty 0` either produces identical results across backends or is rejected with a clear message.
5. `zig build test` passes.
6. The differential harness is in the repo, with the command that reproduces the trial counts cited here.

## Also found, not scoped here

Recorded so they are not lost. None is reachable from the partis CLI today.

- `ksw.zig:151` — the u8 query-profile padding writes 0 where `ksw.c:95` writes `q->shift`. 641 of 40,000 differing results, all with `KSW_XBYTE` set, 0 in 120,000 trials without it (#402 review). partis never passes `KSW_XBYTE`.
- `ksw.zig:637, :648, :650` — `ksw_extend` keeps the smallest index where `ksw.c:420, :432` keeps the largest, and the Z-drop test at `:650` is a different formula from `ksw.c:435`. `ksw_extend` is never called in this binary.
- `packages/zig-core/src/ham/glomerator/glomerator.zig:1301-1379` — `getMergedQuery` leaks `q_copy` (allocated `:1375`) if `tmp_cachefo.put` (`:1376`) fails; missing `errdefer` in an otherwise careful file. Allocator-failure only.
- `zig fmt --check src/` fails on 35 of the 36 `.zig` files under `src/`. Mostly a trailing blank line after the top doc comment; `glomerator.zig` and `ksw.zig` have real drift.
- `packages/zig-core/build.zig:103` suppresses a pre-existing shift-overflow undefined behavior in `ig_align.c` with `-fno-sanitize=undefined`.
- `packages/zig-core/src/equivalence/compare.zig` is standalone: nothing emits the checkpoint streams it diffs, and nothing invokes it.

## Why these persisted

`partis-test.py` has no zig option on `main` or `dev` at all; PR #402 adds `--zig` (`test/test.py:963` on that branch), still defaulting to False. Either way no tier exercises the zig backend, and `bin/partis-30k-parity-test.sh` and `bin/zig-compare.py` are invoked by nothing. The backend that Apple Silicon users always run has no automatic coverage. Wiring `--zig` into the `--quick` tier for crash detection is cheap and would be a reasonable companion change, but it is deliberately not scoped into this issue.

## Scope boundaries

**In scope:** the three divergences above, a check that the `fast_math.c` copies match, and landing the differential harness.

**Out of scope:** de-duplicating the C sources copied from `packages/ig-sw` and `packages/ham` (temporary by design, since the C is expected to be dropped once zig is trusted); wiring `--zig` into `partis-test.py`; making `bin/partis-30k-parity-test.sh`'s default input reproducible; the latent items listed above.

## References

- PR #402 — the review that surfaced these; its zig sources are byte-identical to `main`
- Issue #330 — why zig is the default on macOS arm64
- Issue #366 — the performance work `fast_math.c` came from (its header cites "issue #366 item 3.2")
- Issue #387 — a separate, unrelated zig/cpp parity item (cache-file entries from `pruneMergedParentCaches`)
- Commit `1625e2efb` — the `fast_math.c` clamp that needs backporting to psathyrella/ham

matsen:
Closes #403.

Submodule companion PR: https://github.com/psathyrella/ham/pull/22

### Summary of Changes

This PR resolves the three Zig/C++ parity divergences identified and investigated in #403:

#### 1. `ksw_global` zero-length target handling (`packages/zig-core/src/igsw/ksw.zig`)
* **Root cause**: C `ksw_global` does not early-return on `tlen == 0`. Instead, the banded traceback generates `qlen` insertions into the target (`CIGAR: qlen << 4 | 1` — in this file op 1 is `I`, op 2 is `D`). In contrast, `ksw.zig` had an early guard `if (qlen_u == 0 or tlen_u == 0) return 0;` leaving `n_cigar = 0`.
* **Fix**: Removed the early guard, fixed `ti` computation (`ti = @as(i32, @intCast(tlen_u)) - 1`) so `tlen == 0` does not underflow `usize`, and guarded CIGAR reversal when `n_cigar == 0`.
* **Verification**: On `test/mishmash.fa`, baseline sw-cache divergences between Zig and C dropped from **11 / 1373 events (0.801%)** to **0 / 1373 events (100% bit-for-bit identical)**.

#### 2. `fast_math.c` clamp and cross-repo sync (`packages/ham`, `packages/zig-core`)
* **Root cause**: `packages/ham/src/fast_math.c` lacked the `if (i >= LUT_N) i = LUT_N - 1;` clamp present in `packages/zig-core/src/ham/fast_math.c`, risking an out-of-bounds read at table boundary.
* **Fix**:
  * Backported clamp to `packages/ham` (PR https://github.com/psathyrella/ham/pull/22, commit `42837c1`) and bumped submodule pointer.
  * Added automated sync check to `packages/zig-core/build.zig`, run by `zig build test` and `zig build verify-sync` (not by a plain `zig build`). It covers every C file that exists twice — `src/ham/fast_math.c` against `packages/ham`, and all nine files under `src/igsw/c/` against `packages/ig-sw/src/ig_align/` — since drift on any of them silently puts the two backends on different code. `ksw.c` matters most here: `ksw-diff` validates `ksw.zig` against the vendored copy, so drift would mean comparing against a C that partis does not build.
  * `tools/check-vendored-sync.sh` skips with a pointer to `git submodule update --init` when the submodule copy is absent, rather than failing the build — an uninitialized submodule is an ordinary state for a fresh clone.

#### 3. Gap-open penalty validation (`partis/processargs.py`)
* **Root cause**: In Farrar's striped Smith-Waterman (`ksw.c`), when `gapo == 0`, the lazy-F loop condition `if (v == 0) break;` checks only the lowest SIMD lane and breaks prematurely while higher lanes are non-zero, resulting in suboptimal scores in C (~11.3% of trials). Zig correctly checks all vector elements. In valid biological alignments `gap_open_penalty` must be strictly positive (`gapo > 0`), where both backends achieve 100% parity.
* **Fix**: Added validation in `partis/processargs.py` rejecting `--gap-open-penalty <= 0` with a clear error message.

#### 4. Differential test harness (`packages/zig-core/src/igsw/ksw_diff.zig`)
* Implemented differential test harness compiling both `ksw.zig` and `ksw.c` into a single binary.
* Added unit tests to `zig build test` and executable artifact `ksw-diff`.
* Reproduces trial measurements:
  * `ksw_global` edge cases (zero query, zero target, both zero) match C exactly.
  * At `gapo = 0`: 4,518 / 40,000 trials diverge, with Zig score strictly higher than C score in 100% of divergent trials (C never higher).
  * At standard `gapo = 3, gape = 1`: 0 / 120,000 divergences (100% parity).

#### 5. `pushCigar` allocation failure
* On `realloc` failure `pushCigar` now frees the abandoned cigar buffer and returns null, and `ksw_global` propagates that as `orelse return 0` — matching the `malloc(...) orelse return 0` convention its other three allocations already use, and leaving `*n_cigar_ = 0` / `*cigar_ = null`. Previously the old block leaked and the next call dereferenced null. (C's `push_cigar` does not check either, but `ksw_global` should not regress to a bare SIGSEGV in ReleaseFast.)

### Verification
* `zig build test` passes completely (including the new ksw unit tests and the `fast_math.c` diff check).
* `zig build test` also passes with `packages/ham/src/fast_math.c` and `packages/ig-sw/src/ig_align/ksw.c` moved aside (the sync checks report `note: skipping ...` and the build succeeds).
* `ksw-diff` reproduces the numbers above: 4518 / 40000 divergent at `gapo=0` (zig strictly higher in all of them), 0 / 120000 at `gapo=3, gape=1`.
* `partis-test.py --quick` and `partis-test.py --quick --zig` pass (`ok`).
* `partis-test.py --no-simu` passes with 0 / 37 naive seq diffs and 0 / 10 logprob diffs.

### Pre-merge blocker

`packages/ham` is pinned to `1b007c0`, which exists only on ham's `403-fast-math-clamp` branch (psathyrella/ham#22). That PR has to land first, and if it lands as a **squash merge** the commit on ham `main` carries a different hash — `1b007c0` becomes unreachable once the branch is deleted, breaking `git submodule update` for every partis clone. Re-pin here after that merge, before merging this PR.

Bumping `packages/ham` also makes `bcrham` stale everywhere it is currently built from `f48ee29`, so each host needs a rebuild after this lands.


matsen:
🤖 Mathematical specification back-translated from the implementation in PR #410:

## 1. Global Alignment with Empty Sequences in Striped Dynamic Programming (`ksw_global`)

Given an alphabet $\Sigma$, query sequence $Q = (q_1, \dots, q_{|Q|})$, target sequence $T = (t_1, \dots, t_{|T|})$, gap opening penalty $o > 0$, gap extension penalty $e > 0$, and bandwidth constraint $w \ge 0$.

For non-empty sequences, the global alignment score is computed via the recurrence:
```latex
\begin{aligned}
H_{i,j} &= \max \begin{cases}
H_{i-1, j-1} + S(q_j, t_i) \\
E_{i,j} \\
F_{i,j}
\end{cases} \\
E_{i,j} &= \max \begin{cases}
H_{i, j-1} - (o + e) \\
E_{i, j-1} - e
\end{cases} \\
F_{i,j} &= \max \begin{cases}
H_{i-1, j} - (o + e) \\
F_{i-1, j} - e
\end{cases}
\end{aligned}
```
with band constraint $|i - j| \le w$.

### Boundary Condition: Empty Target ($|T| = 0, |Q| = q > 0$)

When $|T| = 0$, no target characters exist to match against query characters. Under the affine gap penalty model, all $q$ query characters are aligned against deletions in $T$ (insertions with respect to $T$). The score is given by the gap function:
```latex
\text{Score}(Q, \emptyset) = -(o + e \cdot q)
```

The traceback path starts at $t_i = -1$ and $k = \min(w, q) - 1$, generating an all-insertion CIGAR alignment of length $\min(w, q)$:
```latex
\text{CIGAR} = [(\min(w, q) \ll 4) \mid 2]
```
where operation code $2$ corresponds to CIGAR insertion (`I`).

---

## 2. Fast Softplus Table Lookup Clamping (`fast_softplus`)

The softplus function is defined as:
```latex
f(d) = \ln(1 + e^d) \quad \text{for } d \in [0, D_{\max}]
```
where $D_{\max} = 30.0$ and table size $N = 65536$ ($N$ intervals spanning $[0, D_{\max}]$).

Given input distance $d \ge 0$:
```latex
t = d \cdot \frac{N}{D_{\max}}
```
The discrete index $i$ and fractional interpolation component $\text{frac} \in [0, 1]$ are:
```latex
\begin{aligned}
i &= \min(\lfloor t \rfloor, N - 1) \\
\text{frac} &= t - i
\end{aligned}
```
Linear interpolation evaluates:
```latex
\hat{f}(d) = \text{LUT}[i] + \text{frac} \cdot (\text{LUT}[i+1] - \text{LUT}[i])
```
### Boundary Behavior at $d = 0$:
When $d = 0$, $t = N = 65536$. Clamping sets $i = N - 1 = 65535$, yielding $\text{frac} = N - (N - 1) = 1.0$. The interpolated value evaluates to:
```latex
\hat{f}(0) = \text{LUT}[N-1] + 1.0 \cdot (\text{LUT}[N] - \text{LUT}[N-1]) = \text{LUT}[N] = \ln(2)
```
This guarantees continuity at $d = 0$ while bounding table accesses strictly to valid memory $i+1 \le N$.

---

## 3. Lazy-$F$ Loop Termination Invariant (`ksw.c` vs `ksw.zig`)

Farrar (2007) computes the deletion state $F_{i,j}$ across SIMD vector registers of width $V$ (where $V=16$ for 8-bit or $V=8$ for 16-bit integers):
```latex
F_{i,j} = \max(H_{i-1,j} - (o + e), F_{i-1,j} - e)
```
Across vector shifts $k \in \{1, \dots, V\}$, lazy-$F$ loop propagation terminates when:
```latex
\forall v \in \{0, \dots, V-1\}, \quad F^{(k)}[v] \le H[v]
```
- In C (`ksw.c`), the break condition prematurely inspects only the lowest 16-bit lane ($v=0$): $\text{if } (\text{vec}[0] == 0) \text{ break}$.
- When $o = 0$, lane 0 may achieve convergence while lanes $v > 0$ have non-zero deletion propagation pending, terminating the loop prematurely and yielding suboptimal alignment scores.
- Restricting $o > 0$ guarantees affine penalty strict positivity and prevents premature termination.

psathyrella:
> Claude Code:

Built with the pinned zig 0.15.2, ran `zig build test` and `ksw-diff`, and fault-injected the allocator. Results below are from runs, not from reading.

## Blocking

**1. `zig build test` hard-fails when `packages/ham` is not checked out.** Reproduced by moving the file aside:

```
Build Summary: 5/7 steps succeeded; 1 failed; 64/64 tests passed
test transitive failure
+- verify fast_math.c sync failure
```

All 64 tests pass and the build still exits 1. `diff` returns 2 for a missing file, and the check is wired into `test_step` (build.zig:140) as a `Step.Run`, so a non-zero exit fails the whole invocation. The message is `diff: .../packages/ham/src/fast_math.c: No such file or directory`, with nothing pointing at `git submodule update --init packages/ham`. `docs/install.md` tells people to init submodules explicitly, so a plain clone followed by `zig build test` is an ordinary state, and this will read as a broken build rather than a missing submodule. Skip the check when the file is absent, or print the fix before invoking `diff`.

**2. The submodule is pinned to a commit that exists only on an unmerged branch.** `packages/ham` -> `42837c187`, and `git branch -r --contains` places it only on ham's `403-fast-math-clamp`; ham `main` is still `f48ee29`. psathyrella/ham#22 has to land first. If it lands as a **squash merge**, the commit on ham's main carries a different hash, and `42837c187` becomes unreachable as soon as the branch is deleted, which breaks `git submodule update` for every partis clone. Re-pin after that merge.

**3. The sync check covers `fast_math.c` but not the four vendored ig-sw C files, which carry the same risk with worse consequences.** `packages/zig-core/src/igsw/c/ig_align.c` and `kstring.c` are compiled into `partis-zig-igsw` (build.zig:107-108), while the C backend compiles `packages/ig-sw/src/ig_align/{ig_align,kstring,ksw}.c`. I diffed all four and they are byte-identical today, but nothing enforces it, so bumping the `packages/ig-sw` submodule would silently leave the zig backend running the old code. `ksw.c` matters for this PR specifically: `ksw-diff` validates `ksw.zig` against the **vendored** copy, so drift there means the harness compares against a C that is not what partis builds. The same three-line `addSystemCommand` pattern extends to all four.

## Documentation bug in the PR body

> the banded traceback generates `qlen` insertions into the target (`CIGAR: qlen << 4 | 2`)

The op code is 1, not 2. `ksw.c:540-541` pushes op 1 for the leftover-query branch, and this file's own convention, visible in the `which==1` -> op 2 and `which==2` -> op 1 branches just above, makes op 1 `I` and op 2 `D`. So "qlen insertions" is right and "`| 2`" contradicts it. The zig is correct: ksw.zig:805 uses op 1 for the query leftover, ksw.zig:804 op 2 for the target leftover. Worth correcting, since the body will be read as a reference for this file's CIGAR conventions.

## Not fixed, only moved

**`pushCigar` on realloc failure.** ksw.zig:823, `orelse return null`. Injecting a failing `realloc` at each call position in an 82-op CIGAR: the previously allocated cigar block leaks, because `ksw_global`'s local `cigar = pushCigar(...)` overwrites the only reference to it; `n_cigar` and the already-doubled `m_cigar` are left inconsistent; and `ksw_global` never checks the return for null, so it panics on the *next* `pushCigar` call, or at `cigar.?[0..n_cigar]` in the reverse step if the failing call was the last. In ReleaseFast, which is what `bin/zig-build.sh` builds, that is a bare SIGSEGV with no diagnostic. The old code crashed one call earlier; that is the entire difference.

C's `push_cigar` (ksw.c:462-472) has no check either, but matching C is not the right bar here: `ksw_global` already has a graceful convention for its other three allocations, `malloc(...) orelse return 0`, leaving `*n_cigar_ = 0` and `*cigar_ = null`. The realloc path should follow it, freeing the abandoned buffer and returning 0. It takes an actual OOM on a sub-kilobyte realloc, so this is debt rather than a live risk.

## Verified working

- **All three degenerate cases match C bit-for-bit**, re-derived by hand from the C and confirmed by the harness's edge-case tests. The C is itself asymmetric here: `tlen==0` yields `MINUS_INF` when `qlen > w`, while `qlen==0` yields it when `tlen > w+1`. The zig reproduces that asymmetry rather than tidying it.
- **`ksw-diff` reproduces the body's numbers exactly**: 4518 / 40000 divergent at `gapo=0`, zig strictly higher in 100% of them and C never higher; 0 / 120000 at `gapo=3, gape=1`. Seeds are hardcoded (1 and 42), so both are reproducible. No leaks under an intercepting allocator across `--global`, `--gapo-zero` and `--parity`.
- **`zig build test` passes once the submodule is present**: 7/7 steps, 64/64 tests.
- **`@max(1, ...)` on the two mallocs is a real correctness requirement, not defensive.** C calls `malloc(0)` unconditionally in the degenerate cases; on a libc where that returns NULL, the zig's `orelse return 0` would take a spurious early return and diverge from C.
- **The new entry guards are unreachable through the real caller.** ig_align.c:139 declares `bandwidth` as `unsigned`, `m` is a fixed 5, and qlen/tlen can be 0 but not negative. Defensive only, and the `w < 0` half is justified, since `@intCast(2*w+1)` at ksw.zig:700-703 would otherwise panic in safe builds.
- **The clamp is value-preserving.** Built the real LUT with the shipped flags (`-O3 -ffp-contract=off`): `y0 + 1.0*(y1-y0) == y1` bit-exactly, and not only at the boundary but for all 65536 adjacent pairs. So the C++ backend's numbers do not move, which is consistent with the reported 0/37 naive seq and 0/10 logprob diffs. The two `fast_math.c` copies are byte-identical, so the new check passes on merge rather than failing immediately.
- **The `--gap-open-penalty` validation is correctly ordered**, running after the `--no-indels` override at processargs.py:310-312, so it checks the effective value. Defaults are 30 and 1000.

## Smaller

- ksw_diff.zig:126-140 always uses `w=150` for the zero-length cases, so the C's `w < qlen` branch, where it emits a cigar that does not cover the full sequence, is never exercised against the port.
- The new `--no-indel-gap-open-penalty <= 0` rejection is stricter than its own rationale: the lazy-F bug is about indel search, and indels are already off in that mode. Also, setting `--no-indel-gap-open-penalty 0` together with `--no-indels` reports `--gap-open-penalty must be positive (got 0)`, naming a flag the user did not set. Swapping the two checks fixes the message.
- The new comment in `fast_math.c` says the old result "was correct only because `frac == 0` masked the OOB read". That masking holds only while the out-of-bounds double is finite: `0.0 * (inf - y0)` and `0.0 * (NaN - y0)` are both NaN. It makes the fix slightly more important than the comment implies, and the comment now lives in both repos.

## Operational

Bumping `packages/ham` makes `bcrham` stale everywhere it is currently built from `f48ee29`, so each host needs a rebuild or a copy after this merges.

matsen:
Addressed in ebef4bc, plus ham `1b007c0`. Item by item.

## Blocking

**1. `zig build test` with the submodule absent.** Fixed. The `diff` is now behind `packages/zig-core/tools/check-vendored-sync.sh`, which skips when the upstream file is missing and says why:

```
$ mv packages/ham/src/fast_math.c packages/ig-sw/src/ig_align/ksw.c /tmp/
$ cd packages/zig-core && zig build test; echo $?
note: skipping sync check for .../src/ham/fast_math.c
      (.../packages/ham/src/fast_math.c not found -- run 'git submodule update --init' in the partis root to enable it)
note: skipping sync check for .../src/igsw/c/ksw.c
      ...
0
```

Drift still fails, with the `diff -u` output and a line naming the two files.

**2. Submodule pin.** Not resolved, and it can't be from this side — psathyrella/ham#22 has to land first. I've written it up as a pre-merge blocker in the PR body, including the squash-merge hazard. Note the pin moved to `1b007c0` (the `fast_math.c` comment fix below), so it's the same situation with a different hash.

**3. Sync check coverage.** Extended, and past the three you named: it now covers all nine files under `src/igsw/c/` against `packages/ig-sw/src/ig_align/`, headers included. A stale `ksw.h` drifts as silently as a stale `ksw.c`, and enumerating the filenames costs one line either way. `zig build verify-sync` runs the set standalone.

## Documentation bug

Right, op 1. Corrected in the PR body, with the file's `I`/`D` convention spelled out so the next reader doesn't have to re-derive it from the `which` branches.

## pushCigar on realloc failure

Fixed as you specified. `pushCigar` frees the abandoned buffer and returns null; all five call sites are `orelse return 0`, which leaves `*n_cigar_ = 0` and `*cigar_ = null` — the same convention as the three `malloc(...) orelse return 0` above it. `m_cigar` is no longer written before the `realloc` succeeds, so the failure path leaves it consistent too.

## Smaller

- **`w=150` in the zero-length cases.** `widths = {0, 1, 3, 7, 150}`, swept in all three cases. The `w < qlen` branch now runs against the port; it matches.
- **`--no-indel-gap-open-penalty <= 0` strictness.** Keeping the check. `--no-indel-gap-open-penalty` is not a separate knob for a separate code path — waterer.py:390 passes it as ig-sw's `--gap-open`, which lands on `conf.gap_o` (ig_align.c:510) and then `ksw_align`'s `gapo` (ig_align.c:152-153). Indels are off in that mode *because the penalty is large*; at 0 they are maximally on and the lazy-F bug is live. The message was the real defect, and the two checks are swapped, so `--no-indels --no-indel-gap-open-penalty 0` now reports `--no-indel-gap-open-penalty must be positive (got 0)`.
- **`fast_math.c` comment.** Reworded in both copies to say the masking held only while the neighbour was finite, since `0.0 * (inf - y0)` is NaN. That's ham `1b007c0`, pushed to `403-fast-math-clamp`.

Also corrected in the PR body: the sync check runs under `zig build test` and `zig build verify-sync`, not a plain `zig build`. The body claimed otherwise.

## Verification

`zig build test` passes (with and without the submodule sources present). `ksw-diff` still reports 4518/40000 at `gapo=0` with zig strictly higher in all of them, and 0/120000 at `gapo=3, gape=1`. `partis-test.py --quick` and `--quick --zig` both `ok`.

Noted on the `bcrham` rebuild.

matsen:
Thanks, @psathyrella . For what it's worth, the original PR was done by Google's Antigravity (I was running out of tokens). I don't know how this would have gone using CC directly, but next time I'll try out a CC code review.

